Issue #6, opened 2020-05-20
Output IRF format


HealthyPear:
**Introduction:**

This topic is still in an unofficial state and will likely evolve in time.
To check on the current its status, check the relevant sub-tasks of the IRF-WG [here](https://forge.in2p3.fr/projects/instrument-response-functions/wiki#Task-Assignment).

Feel free to add or link here any related issue / proposal / discussion.

**Current status:**

It is somewhat clear that we should follow the format described by the [Gamma Astro Data Format (GADF)](https://gamma-astro-data-formats.readthedocs.io/en/latest/).
We are following the latest development version.

MaxNoe:
I think we should just make the decision, that we want to use OGADF and when we find it lacking something CTA requires, we will include it there.

After all, the intended purpose of OGADF is to become the CTA format and provide documentation of that format to the community.

HealthyPear:
> I think we should just make the decision, that we want to use OGADF and when we find it lacking something CTA requires, we will include it there.
> 

This is indeed the current "decision" we took

MaxNoe:
Ok, then let's close this for now and reopen if this needs discussion again.

HealthyPear:
Well, I actually left it open for this purpose, since we will have soon people working on this...

HealthyPear:
I mean this issue was not a bug or a one-time issue, it was meant as an open discussion to update along the development

HealthyPear:
I would reopen it light of this

MaxNoe:
I would rather see concrete issues as soon as something comes up.

E.g. an issue here when pyirf is violating some part of the standard or an issue at OGADF when something new is needed there. 

Having many open issues where we don't really expect any discussion is not a good idea as it will bury the issues that actually need work.

Especially with an eye on the coding sprint, we should aim at having concrete issues that contributors can start to tackle.

MaxNoe:
Again, if there is currently nothing that needs discussion here, leave it closed. As soon as something comes up, reopen or open a new, more concrete issue.

But for now, the question has an answer.

HealthyPear:
We should decide how to handle all auxiliary data (needed at least for diagnostics/testing) that can be produced together with the actual output file that we have to provide (basically the single FITS file containing the IRFs).

This data should be, e.g:
- the final selection cuts used to produce the IRFs (part of the provenance data of _pyirf_ - see #39)
- modified versions of some IRFs (e.g. collection areas with part of or no cuts applied)
- ...

[...]

PR #8, opened 2020-05-28
Start refactoring aeff, edisp and init of IRFMaker

JouvinLea:
This PR remove the dependency of IRFMaker on the table obtained from cut optimisation.
This is useless because it is used only to loop on the reco energy bin and those are already in the config file. It is also used in make_bkg to compute the acceptance ratio between gamma and hadron but now instead the method of make_bkg take as an argument the angular cut for each reco bin.

It also adds again the event dict parameter to initialize the IRFMaker. 

It makes make_aeff DL3 format compatible so it adds an offset dependency for the effective area. In the future for simulation with Diffuse MC, we will take it into account properly. 
For the effective area, it stores now in the hdu the 2D array (Noffset, Ntrue) and for the energy dispersion it stores now the 3D array: (Noffset, Nmigra, Ntrue)

### DIFF
--- a/pyirf/perf/irf_maker.py
+++ b/pyirf/perf/irf_maker.py
@@ -163,6 +163,9 @@ class IrfMaker(object):
     ----------
     config: `dict`
         Configuration file
+    evt_dict : `dict`
+        Dict for each particle type, containing a table with the required column for IRF computing.
+        TODO: define explicitely the name it expects.
     outdir: `str`
         Output directory where analysis results is saved
     """
@@ -173,32 +176,24 @@ class IrfMaker(object):
 
         # Read data saved on disk
         self.evt_dict = {}
-        for particle in ["gamma", "electron", "proton"]:
-            self.evt_dict[particle] = pd.read_hdf(
-                os.path.join(outdir, "{}_processed.h5".format(particle))
-            )
+        #Loop on the particle type
+        for particle in evt_dict.keys():
+            self.evt_dict[particle] = evt_dict[particle]
 
-        # Read table with cuts
-        self.table = Table.read(
-            os.path.join(
-                outdir, "{}.fits".format(config["general"]["output_table_name"])
-            ),
-            format="fits",
-        )
-        self.table = self.table[np.where(self.table["keep"].data)[0]]
-
+        cfg_binning_ereco = config["analysis"]["ereco_binning"]
+        cfg_binning_etrue = config["analysis"]["etrue_binning"]
+        self.nbin_ereco = cfg_binning_ereco["nbin"]
+        self.nbin_etrue = cfg_binning_etrue["nbin"]
         # Binning
         self.ereco = np.logspace(
-            np.log10(self.table["emin"][0]),
-            np.log10(self.table["emax"][-1]),
-            len(self.table),
+            np.log10(cfg_binning_ereco["emin"]),
+            np.log10(cfg_binning_ereco["emax"]),
+            self.nbin_ereco+ 1,
         )
-
-        cfg_binning = config["analysis"]["etrue_binning"]
         self.etrue = np.logspace(
-            np.log10(cfg_binning["emin"]),
-            np.log10(cfg_binning["emax"]),
-            cfg_binning["nbin"] + 1,
+            np.log10(cfg_binning_etrue["emin"]),
+            np.log10(cfg_binning_etrue["emax"]),
+            self.nbin_etrue + 1,
         )
 
     def build_irf(self):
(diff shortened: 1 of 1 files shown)


pyirf/perf/irf_maker.py
```diff
@@ -173,32 +176,24 @@ def __init__(self, config, evt_dict, outdir):
...
-        self.table = Table.read(
-            os.path.join(
-                outdir, "{}.fits".format(config["general"]["output_table_name"])
-            ),
-            format="fits",
-        )
-        self.table = self.table[np.where(self.table["keep"].data)[0]]
-
+        #Loop on the particle type
+        for particle in evt_dict.keys():
```
MaxNoe, line 180: why not `self.evt_dict = evt_dict`?

pyirf/perf/irf_maker.py
```diff
@@ -409,19 +411,55 @@ def make_effective_area(
...
+        table_energy["ETRUE_HI"] = Column(
+            energy_true_hi,
+            unit="TeV",
+            description="energy max",
+            format=str(len(energy_true_hi)) + "E",
         )
-        t[hdu_name] = Column(area, unit="m2", description="Effective area", format="E")
 
-        return IrfMaker._make_hdu(hdu_name, t, ["ENERG_LO", "ENERG_HI", hdu_name])
+        # Needed for format, a bit hacky...
```
MaxNoe, line 431: Why these two intervals?

MaxNoe: I f you are doing point-like IRFs, theta bins should have exactly one entry which is the distance of the on region to the camera center (0 for on, wobble distance for wobble observations).

pyirf/perf/irf_maker.py
```diff
@@ -523,6 +561,58 @@ def _make_hdu(cls, hdu_name, t, cols):
         hdu.header.set("EXTNAME", hdu_name)
         return hdu
 
+    @classmethod
+    def _make_aeff_hdu(cls, table_energy, table_theta, aeff):
+        """List of columns"""
```
MaxNoe, line 566: See https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/blob/master/source/irfs/full_enclosure/aeff/create_example_file.py

for a more concise way of creating these BinaryHDUs if you already have astropy tables.

pyirf/perf/irf_maker.py
```diff
@@ -523,6 +561,58 @@ def _make_hdu(cls, hdu_name, t, cols):
...
+                    "AEFF",
+                    aeff["AEFF"].format,
+                    unit=aeff["AEFF"].unit.to_string(),
+                    array=np.expand_dims(aeff["AEFF"].data, 0),
+                    dim=str(aeff["AEFF"].data.T.shape),
+                ),
+            ]
+        )
+
+        hdu.header.set(
```
MaxNoe, line 603: You should not be needing to set this yourself, astropy automatically takes care of columns types and shapes.

Why do you set this?

pyirf/perf/irf_maker.py
```diff
@@ -523,6 +561,58 @@ def _make_hdu(cls, hdu_name, t, cols):
...
+            "TDIM7",
+            "("
+            + str(len(table_energy["ETRUE_LO"]))
+            + ","
+            + str(len(table_theta["THETA_LO"]))
+            + ")",
+        )
+        hdu.header.set(
+            "EXTNAME", "AEFF", "name of this binary table extension "
```
vuillaut, line 612: The recommended EXTNAME keyword is “EFFECTIVE AREA”.
from https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/aeff/index.html

pyirf/perf/irf_maker.py
vuillaut: No specific reason for these values but a range is required for gammapy to work even though for now we are only working with point source MC.
We'll need to make a cleaner version later to work on diffuse.
Maybe add more comments here @JouvinLea ?

pyirf/perf/irf_maker.py
vuillaut: It looks cleaner indeed.
If you can address that comment @JouvinLea that'd be great.

JouvinLea:
I took all the comments I think.
Of course then a lot of cleaning is still needed in another PR.
1) PSF code is not DL3 compatible at the moment
2) have one function to create the HDUlist
3) Add unit test
4) Create offset dependant IRF with diffuse MC, add point-like or full enclosure IRF

this will be done in another PR but it will be great to be able to merge this one that one can start working on edisp and effective area point-like computation

pyirf/perf/irf_maker.py
HealthyPear: The specific nomenclature depends on the version of the GADF.
We should decide which one do we want to work with.
As a preliminary thought, in issue #6 I proposed to use the latest stable release instead of the very latest development one.
In that case this name would be instead `EFFECTIVE AREA (RECO)`.

JouvinLea: ok but on my new commit this is already change.....

HealthyPear: sure no problem!
I just wanted to recall that we didn't fix this choice for now (if for some reason it is obvious and I missed it!)
In any case, this is a detail about the output format, so it is not critical for now.

JouvinLea: yes I agree!!!

vuillaut: So we can AEFF for now?

Issue #3, opened 2020-05-20
Maintenance


HealthyPear:
This is a **summary issue**.

The following a list of things that should be started to aid/foster development and usage.
Please, open/work on separate issues and link them here.

- [x]  Continuous integration (CI)
- [x] Unit-tests
- [ ] Test data
   - [ ] Diffuse gamma test data (#54)
   - [ ] Smaller test data set for unit tests (#55)
- [x] Integration tests (#24)
- [x] Documentation
- [x] Benchmarks (#36)
- [x] Continuous deplyoment (CD)
  - [x] #42 
  - [x] update Travis configuration file

In the future, depending on the complexity of this software, this summary issue could become a project by itself.

maxnoe:
I am closing this. We should use labeling to group issues, not master issues that stay open forever.

HealthyPear:
we didn't finish this, we are still missing the diffuse test data

what 's wrong with issue? it is a summary of what is needed with link to the corresponding issues

maxnoe:
Like I said, I think we should use the proper feature of github to group tests: labels.

So instead of an issue Named Maintenance, that is just a list of other issues, we have a label maintenance and github can aautomatically show us all issues with label maintenance.

PR #9, opened 2020-05-29
Start setup of CI

HealthyPear:
This PR, related to #3 , adds a basic Travis CI configuration file.

The CI for the moment takes care only for installation of the environment and of _pyirf_ itself, but the rest (pytest, codecov, and building of the docs) can be simply un-commented as soon as the repository is set up properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,49 @@
+language: generic
+
+env:
+    global:
+        - PYTHONIOENCODING=UTF8
+        - MPLBACKEND=Agg
+
+matrix:
+  include:
+    - os: linux
+      language: generic
+      env:
+        - PYTHON_VERSION=3.6
+        - CONDA=true
+
+    - os: linux
+      language: generic
+      env:
+        - PYTHON_VERSION=3.7
+        - CONDA=true
+
+before_install:
+
+    - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    - bash miniconda.sh -b -p $HOME/miniconda
+    - . $HOME/miniconda/etc/profile.d/conda.sh
+    - hash -r
+    - conda config --set always_yes yes --set changeps1 no
+    - conda update -q conda  # get latest conda version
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+    - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    - bash miniconda.sh -b -p $HOME/miniconda
+    - . $HOME/miniconda/etc/profile.d/conda.sh
+    - hash -r
+    - conda config --set always_yes yes --set changeps1 no
+    - conda update -q conda  # get latest conda version
+    - conda info -a # Useful for debugging any issues with conda
+
+install:
+    - conda env create -f environment.yml
```
MaxNoe, line 33: This suffers from the same problem as lstchain and ctaplots: it does not use the python version you give in the `PYTHON_VERSION` environment variable.

I would suggest to just not use conda (testing with a pure pip install is fine and actually more general) or use the solution employed by ctapipe, lstchain and others, namely to replace the python version in environment.yml with sed

.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+    - conda activate pyirf
+    - pip install travis-sphinx codecov pytest-cov
+    - python setup.py install
+
+script:
+    # - pytest --cov=protopipe
+    # - travis-sphinx -v --outdir=docs/_build build -n --source=docs/
+
+after_script:
```
MaxNoe, line 42: Why do you need to deactivate conda? Why the if if you always use conda?

HealthyPear: This is just if one wants to add other tools later on (I always used conda, but we can add other things if you think they are better )

.travis.yml
HealthyPear: I never actually realized this.
if it's not super urgent (namely using 3.7 or 3.8 - I guess it will use always the latest stable release that comes with conda base?) maybe we can fix this in another PR?
I would suggest to fix things generally for the moment, so we know that there is a base backup always working

Issue #7, opened 2020-05-20
Reproduce current on-axis EventDisplay IRFs


HealthyPear:
This will be the first real test!

For simplicity, I would propose to stick to on-axis for 2 reasons:
- baby steps
- the _protopipe_ code from which _pyirf_ was born is already tailored for _best sentivity point-source analysis_.

In order to have more info on this, we should first be able to manage some first basic input/output to know where we are.

HealthyPear:
We have reached a sufficient level of approximation (also, not counting the statistical error bars that will soon come from #87) so I will close this issue for the time being.

We could now shift our focus on the off-axis analysis, which currently requires us to fix issue #54 .

Issue #10, opened 2020-06-17
Proposal for package structure and refactoring


HealthyPear:
This is an attempt to propose and anticipate how the general structure of the package could look like.

Most of it is already there!

**Code to be removed:**

If needed for current development, it shouldn't be removed immediately, but we could at least move it all into to a separated `plots` folder (see issue #31)

- plotting code (e.g. in `pyirf.scripts.lst_performance`)
- notebooks (there is a `notebooks` folder)

**Package structure:**

- `io`
  - it should contain all input formats - the final one + the temporary ones - see #2
  - we want to provide only one output format, luckily! - see #6 
- `resources`, for example configuration files, perhaps for different analyses settings
- `scripts`, theoretically there should be only one in the end, aka the "tool" to produce the IRFs
- `perf`, could be probably renamed in something more clear and it should contain all the classes to optimize cuts, produce the IRFs (it already the case)

**Notes:**

- part of development is about management of the input and configuration; from the latest release of _ctapipe_ some features are available such as the `ctapipe.core.tool class`; from one side we shouldn't reinvent the wheel, on the other side _pyirf_ is supposed to be an independent tool
- since there should be only 1 input format in the end, perhaps the `io` folder will keep containing only the current `io.py` in which there will be 1 input & 1 output format definition

MaxNoe:
For better testability, it would also be better to disentangle configuration and actually calculating things.

Just one example: in `CutsOptimization`, `compute_weights` could acutally be a free function getting `particle_information` as input instead of the `particle` key and relying on `self.config`. This would make this function much easier to unit test and also remove reliance on the config system from such a low-level function.

HealthyPear:
Further things to do or discuss:

- splitting the main DL3 scripts in e.g.
  - calculate IRFs (maybe even further splitting this in the single IRFs)
  - calculate sensitivity
  - DL3 event list files

bregeon:
> splitting the main DL3 scripts in e.g. 

yes, I was about to suggest that.

maxnoe:
This is largely solved, we have a good, logical project structure I'd say

PR #12, opened 2020-09-04
EventDisplay comparison

### LINKED ISSUE #2: EventDisplay root to DL2 hdf/fits converter
As a first step to get nice DL2 data to test this we could start
by writing a conversion tool from the DL2 data Gernot provided to an hdf5 based DL2
format similar to the DL1 format now in ctapipe and/or to the ogadf fits format.

### LINKED ISSUE #14: Bug in cut optimization
In the following code checking the particle stats (line 387), the condition tested is bugged :
```python
if len(g) <= min_stat or len(g) <= min_stat or len(g) <= min_stat:
                print('Not enough statistics')
                print('  g={}, p={} e={}'.format(len(g), len(p), len(e)))
                key = CutsOptimisation._get_energy_key(emin, emax)
                self.results_dict[key] = {'emin': emin.value, 'emax': emax.value,
                                          'keep': False}
                continue
```
only gamma stats are tested.

HealthyPear:
**Description**

This PR comes with a list of improvements:

- ability to read EventDisplay DL2 data in FITS format,
- prototype of internal data format
- output format started following GADF development version,
- comparison with EventDisplay with "benchmark" notebook (#7 - we should let this open because we miss some optimizations and settings)
- a first unit-test with coverage enabled (part of #3)
- updated documentation with a usage tutorial for the upcoming coding sprint (closes #2)
- bugfix which closes #14


MaxNoe (changes requested):
I made some general comments.

One major point I want to make is that I don't think it is good to include things like comparisons to other irf modules together with the source code of pyirf.

We should keep this repository lean. Maybe another repository for comparisons would be better.

HealthyPear:
> I made some general comments.
> 
> One major point I want to make is that I don't think it is good to include things like comparisons to other irf modules together with the source code of pyirf.
> 
> We should keep this repository lean. Maybe another repository for comparisons would be better.

This comparison is basically an internal (and temporary) benchmark.
For the time being, since _pyirf_ will be used only by us (meaning CTA) I would keep it here to make it easier and faster to test and develop.

One of the points I will make during the next calls is in fact how to host the benchmarking (possible _cta-benchmarks_, but if this is going to be a more general software - meaning not only for CTA - perhaps we will soon open something similar but dedicated to _pyirf_ like what I think you mean)

.prospector.yaml
```diff
@@ -0,0 +1,5 @@
+pep257:
```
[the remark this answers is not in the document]
HealthyPear: Yes, it is to avoid Codacy complaing about things which are a bit annoying and (my opinion) debatable like empty lines sorrounding docstrings (D202, D203) and the fact that in an __init__.py you import but not use (D104)

docs/changelog.rst
```diff
@@ -0,0 +1,20 @@
+.. _changelog:
```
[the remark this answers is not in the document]
HealthyPear: Yes, probably we can merge it...I kept it like this from protopipe where the one in the root directory is listing the versioning and the one in the docs is basically text for introduction (similar to gammapy)

docs/changelog.rst
```diff
@@ -0,0 +1,20 @@
...
+
+This is the changelog for *pyirf*.
+For now we use a one-line description of every pull request, roughly in chronological order.
+
+We don't list every pull request.
+Maintenance and cleanup changes are not of interest to users and are not listed here.
+
+As of today, the people that contributed to *pyirf* or to previous related code are, in alphabetical order by first name:
+
+- Lea Jouvin
```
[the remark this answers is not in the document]
HealthyPear: Yes you're right I didn't think about that

docs/install/index.rst
```diff
@@ -0,0 +1,42 @@
+.. _install:
+
+Installation
+============
+
+The only requirement is an Anaconda (or Miniconda) installation which supports
```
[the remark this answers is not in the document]
HealthyPear: mmm yes it's not a requirement actually

but I think would be still the most common way of using it

I will rewrite that part

docs/install/basic.rst
```diff
@@ -0,0 +1,30 @@
...
+
+  The development version could disrupt functionalities that were working for
+  you, but the latest released version could lack some of those you need.
+
+  To install the latest development version go to :ref:`install-developer`.
+
+If you are a user with no interest in developing *pyirf*, you can start by
+downloading the `latest released version <https://github.com/cta-observatory/pyirf/releases>`__
+
+Steps for installation:
```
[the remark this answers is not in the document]
HealthyPear: OK

docs/conf.py
```diff
@@ -12,54 +12,87 @@
...
-project = 'pyirf'
-copyright = '2020, Julien Lefaucheur, Michele Peresano, Thomas Vuillaume'
-author = 'Julien Lefaucheur, Michele Peresano, Thomas Vuillaume'
+project = "pyirf"
+copyright = "2020, Julien Lefaucheur, Michele Peresano, Thomas Vuillaume"
+author = "Julien Lefaucheur, Michele Peresano, Thomas Vuillaume"
 
 # The full version, including alpha/beta/rc tags
-release = '0.1.0-alpha'
+version = f"{__version__}"
```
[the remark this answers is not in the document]
HealthyPear: OK

docs/changelog.rst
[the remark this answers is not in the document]
HealthyPear: Actually, I forgot I added the AUTHORS file with this PR.

I think it is better to include that file, which is more simple and if some changes happen it doesn't get modified/broken (like e.g. when someone leaves or something with GitHub doesn't work anymore)

It is easy to update it when someone new contributes - even better I add this to the contribution instructions and each author is autonomous

docs/install/index.rst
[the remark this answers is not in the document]
HealthyPear: python3.x with it's pip would be a requirement I guess

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+        type=str,
+        required=True,
+        help="A configuration file like pyirf/resources/performance.yaml .",
+    )
+
+    parser.add_argument(
+        "--obs_time",
+        type=str,
+        required=True,
+        help="An observation time written as (value.unit), e.g. '50.h'",
```
MaxNoe, line 62: That's a really bad idea I think. Just require it to be in hours or separated by a space. Otherwise, what about e.g. half an  hour? `0.5.h`?

MaxNoe: Alternativelive, just let astropy do the parsing. `u.Quantity('50h')` works fine, `u.Quantity('0.5h')

HealthyPear: yep sure, I really don't know why I changed it...I set an empty space between value and unit

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+    )
+
+    args = parser.parse_args()
+
+    # INPUT FROM THE CONFIGURATION FILE
+
+    cfg = load_config(args.config_file)
+
+    # Add obs. time to the configuration file
+    str_obs_time = args.obs_time.split(".")
```
MaxNoe, line 83: see comment above, please just use astropy to parse this: `obs_time = u.Quantity(args.obs_time)`

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+    cfg["analysis"]["obs_time"] = {
+        "value": float(str_obs_time[0]),
```
MaxNoe, line 85: then this becomes `obs_time.value` and `obs_time.unit.to_string('fits')`

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+        "unit": str(str_obs_time[-1]),
```
MaxNoe, line 86: what about `30min`?

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+    # Get input directory
+    indir = cfg["general"]["indir"]
+
+    # Get template of the input file(s)
+    template_input_file = cfg["general"]["template_input_file"]
+
+    # Get output directory
+    outdir = os.path.join(
+        cfg["general"]["outdir"],
+        "irf_{}_Time{}{}".format(  # name could change in future
```
MaxNoe, line 98: This comment seems rather arbitrary, of course this can change.

As a general comments, I would just let users declare the output file name.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+            args.pipeline,
+            # cfg["analysis"]["thsq_opt"]["type"], # not needed now
+            # cfg["analysis"]["obs_time"]["value"],
+            # cfg["analysis"]["obs_time"]["unit"],
+            args.obs_time.split(".")[0],
+            args.obs_time.split(".")[-1],
+        ),
+    )  # and create it if necessary
+    if not os.path.exists(outdir):
```
MaxNoe, line 107: `os.makedirs(outdir, exists_ok=True)`

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # =========================================================================
+    #                   READ DL2 DATA AND STORE IT ACCORDING TO GADF
+    # =========================================================================
+
+    # Load FITS data
+    particles = ["gamma", "electron", "proton"]
+    evt_dict = dict()  # Contain DL2 file for each type of particle
+    for particle in particles:
+        if args.debug:
```
MaxNoe, line 118: See #22

pyirf/scripts/make_DL3.py
HealthyPear: ok, I'll remove the comment

though, if we want them to be compatible it's better we define it - if people shares these files is better they have a unique template no? we can decide later, it is not urgent

MaxNoe: No, we should not rely on filenames for that. Rather we use the corresponding FITS header keywords.

E.g. `CREATOR`, `TELESCOP`, ...

pyirf/scripts/make_DL3.py
HealthyPear: I would do this in a dedicated PR, there is enough material in this one I think (as it is mainly needed to give a runnable script)

pyirf/scripts/make_DL3.py
MaxNoe: This are mostly required by the OGADF standard anyway

pyirf/scripts/make_DL3.py
HealthyPear: also because you already opened an issue about this, so we'll not forget

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
+"""Script to produce DL3 data from DL2 data and a configuration file.
```
MaxNoe, line 1: DL3 data means normally an event list together with the IRFs. What does this script do?

We should probably have a script that calculates irfs, one that calculates sensitivity and one that creates a DL3 event list file from DL2 and irf files.

HealthyPear: it does everything

HealthyPear: > We should probably have a script that calculates irfs, one that calculates sensitivity and one that creates a DL3 event list file from DL2 and irf files.

This can be a refactoring proposal to add to #10

pyirf/scripts/make_DL3.py
HealthyPear: this would have been something to add later to #6 :)

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # Some pipelines could provide some of the DL2 data in different ways
+    # After this part, DL2 data is supposed to be equivalent, regardless
+    # of the original pipeline.
+
+    # Later we should move this out of here, perhaps under a "utils" module.
+
+    if args.pipeline == "EventDisplay":
+
+        # THETA IS SUPPOSED TO BE IN DL2 DATA ACCORDING TO GADF
```
MaxNoe, line 135: No it is not. The only required columns are:

EVENT_ID, TIME, RA, DEC, ENERGY.

Also, OGADF actually only defines the DL3 level, not DL2.

`THETA` is an optional column and is not what you use here. It is the angular distance from the center of the field of view, not from the source position.

https://open-gamma-ray-astro.github.io/gamma-astro-data-formats/events/events.html

HealthyPear: ah ok it is not required but optional

anyway remember that this comes from protopipe in which we work only with on-axis simulations (for now), so the source position is the center of the FoV

HealthyPear: also, probably I am mistaken, but the information on the page you link is indeed also DL2 no? plus maybe something else, but all DL2 information is also there

MaxNoe: So mainly: the comment is wrong. OGADF does not make any statements about simulated DL2 event lists. 
So I would remove any comment that says that the DL2 is not as required as by OGADF and just calculate what is necessary.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+        # EventDisplay provides true and reconstructed directions, so we
+        # calculate it here and we add it to the tables.
+
+        # WARNING:
```
MaxNoe, line 139: Then please directly implement it correctly, what is actually needed here? Also for mono telescopes we also have wobble observations. So there are simulations where we simulated a point source that is not at the center of the field of view.

pyirf/scripts/make_DL3.py
HealthyPear: I modified it

pyirf/scripts/make_DL3.py
MaxNoe: We should just always make the distinction and call the variables accordingly.

pyirf/scripts/make_DL3.py
MaxNoe: Let me make a proposal: we should define our in-memory data structure we need calculate the IRFs.

Then we define reading functions for different formats in the `io` subpackage, e.g. `read_event_display`, which need to return these data structures.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # =========================================================================
+    #                   REST OF THE OPERATIONS (TO BE REFACTORED)
+    # =========================================================================
+
+    # Apply offset cut to proton and electron
+    for particle in ["electron", "proton"]:
+
+        # There seems to be a problem in using pandas from FITS data
+        # ValueError: Big-endian buffer not supported on little-endian compiler
```
MaxNoe, line 167: Yes, you need to convert to native byteorder like so: https://github.com/fact-project/pyfact/blob/65201e6aa029264b0b4305c6bc8731b3a11b79cc/fact/io.py#L127

MaxNoe: but using astropy.table.Table.to_pandas should do that also for you.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+
+        # I convert to astropy table....
+        # should we use only those?
```
MaxNoe, line 170: They better support units but are generally not as powerful when it comes to statistics and other manipulation.

Also tables support ND rows, pandas cannot handle this.

I would see if we can make it work with tables to have the units.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+            np.unique(evt_dict[particle]["OBS_ID"])
+        )
+        cfg["particle_information"][particle]["n_simulated"] = (
+            cfg["particle_information"][particle]["n_files"]
+            * cfg["particle_information"][particle]["n_events_per_file"]
+        )
+
+    # Define model for the particles
+    model_dict = {
+        "gamma": CrabSpectrum("hegra").model,
```
MaxNoe, line 205: This should be configurable. I think also the MAGIC sprectum was used for CTA and especially in the LST region, the difference is quite large.

pyirf/scripts/make_DL3.py
HealthyPear: not sure if we need to modify it...

THETA = angular_separation(
                evt_dict[particle]["TRUE_AZ"],
                evt_dict[particle]["TRUE_ALT"],
                evt_dict[particle]["AZ"],
                evt_dict[particle]["ALT"],
            )

takes already into account off-axis events, am I right?
But in that case, theta 
- is from the FoV center, so the name would be the wrong one,
- I guess TRUE_AZ and TRUE_ALT should be modified in DETX and DETY (I guess they are already in the sky)
But this info is not in the EventDisplay DL2 data

Though, in general, I would leave any off-axis-related modification to a dedicated PR - we are trying to fix the on-axis for the moment.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+            np.log10(cfg_binning["emax"]),
+            cfg_binning["nbin"] + 1,
+        )
+        * u.TeV
+    )
+
+    # Handle theta square cut optimisation
+    # (compute 68 % containment radius PSF if necessary)
+    thsq_opt_type = cfg["analysis"]["thsq_opt"]["type"]
+    if thsq_opt_type in "fixed":
```
MaxNoe, line 224: You should use `==` to compare strings, `in` tests for membership.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+        thsq_values = np.array([cfg["analysis"]["thsq_opt"]["value"]]) * u.deg
+        print("Using fixed theta cut: {}".format(thsq_values))
+    elif thsq_opt_type in "opti":
+        thsq_values = np.arange(0.05, 0.40, 0.01) * u.deg
```
MaxNoe, line 224: The values which are tested for the optimization should also be configurable.

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+                cfg_binning["nbin"] + 1,
+            )
+            * u.TeV
+        )
+        radius = 68
+
+        thsq_values = list()
+
+        # There seems to be a problem in using pandas from FITS data
+        # ValueError: Big-endian buffer not supported on little-endian compiler
```
MaxNoe, line 243: see comment above

pyirf/scripts/make_DL3.py
HealthyPear: DONE

pyirf/scripts/make_DL3.py
HealthyPear: DONE

pyirf/scripts/make_DL3.py
HealthyPear: DONE

pyirf/scripts/make_DL3.py
HealthyPear: DONE

pyirf/scripts/make_DL3.py
```diff
@@ -0,0 +1,338 @@
...
+    # Handle theta square cut optimisation
+    # (compute 68 % containment radius PSF if necessary)
+    thsq_opt_type = cfg["analysis"]["thsq_opt"]["type"]
+    if thsq_opt_type in "fixed":
+        thsq_values = np.array([cfg["analysis"]["thsq_opt"]["value"]]) * u.deg
+        print("Using fixed theta cut: {}".format(thsq_values))
+    elif thsq_opt_type in "opti":
+        thsq_values = np.arange(0.05, 0.40, 0.01) * u.deg
+        print("Optimising theta cut for: {}".format(thsq_values))
+    elif thsq_opt_type in "r68":
```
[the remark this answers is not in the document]
GernotMaier: Angular resolution? No - using 68% containment radius is used (but 80 and 95% quantiles are also filled - useful to get a feeling on the tails)

pyirf/scripts/make_DL3.py
MaxNoe: I don't understand what you are trying to say with your second point. DETX and DETY are something completely different.

pyirf/scripts/make_DL3.py
MaxNoe: > also, probably I am mistaken, but the information on the page you link is indeed also DL2 no? plus maybe something else, but all DL2 information is also there

OGADF only defines the DL3 data format. it makes no statements about anything related to DL2.

pyirf/scripts/make_DL3.py
HealthyPear: maybe you can open an issue about this part

I would like to avoid modifying further this PR (since does already the job) and split the problems in smaller ones.

pyirf/scripts/make_DL3.py
HealthyPear: ah you mean the warning part?

pyirf/scripts/make_DL3.py
HealthyPear: also here, better open an issue to not forget

pyirf/scripts/make_DL3.py
MaxNoe: So again, I would not focus the sprint on making a script. But to create a small set of well-defined functions with a good API that then can be used by anyone to produce IRF components.

pyirf/scripts/make_DL3.py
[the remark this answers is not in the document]
MaxNoe: Maybe it was @moralejo for the MAGIC analysis, I distinctly remember someone saying they use 70% containment for the theta squared cut.

pyirf/scripts/make_DL3.py
MaxNoe: This PR is huge and adds a bunch of code which is not really in good shape. A better approach would be to split this up into a range of smaller PRs. Otherwise we add a bunch of new code we than have to change during the sprint.

HealthyPear: I think the changes we discussed here do not prevent to use the script as it is.
It's already usable by people, partly configurable (e.g. energy, significance).
This was the point of the PR.

I understand that it is not in good shape (in the end as you know it comes from an old part of protopipe), but I think it's better to merge this (which is anyway much better than the previous version - actually the previous version wasn't working at all) and let people use it.

We can lay down the next smaller changes (some of which are already listed here) and take it from there.

pyirf/scripts/make_DL3.py
HealthyPear: also here, better open an issue to not forget

pyirf/scripts/make_DL3.py
MaxNoe: Ok, if this replaces the other scripts, at the very least these have to be deleted. Otherwise we will have three different scripts of which only one is working in the master.

HealthyPear: yes, the old one from protopipe can be deleted (I kept it actually because I was afraid of having forgotten parts of the code, but I can now delete it).

the LST one is something that @vuillaut added and I think it works because there was already a page for LST usage (I then create one for EventDisplay)

MaxNoe: What is the difference? Why are two scripts needed? These duplicate much of the code.

This boils back to the same problem. We should **first** think about data structures and implement small functions before working on these large scripts.

HealthyPear: > What is the difference? Why are two scripts needed? These duplicate much of the code.

I didn't watch into it in detail but from a quick look,
- there is a reader for LSTchain data that we could move to pyirf.io
- plotting code that (in general) in #10 I at least suggest putting under a `pyirf.plot module` (but in this case I see that it comes from ctaplot)
- the main function seems to come from the old script, but I do not know what modifications have been made (I guess not much)

This is something we can solve in a PR for LSTcahin support with @vuillaut 

> This boils back to the same problem. We should **first** think about data structures and implement small functions before working on these large scripts.

Again, I understand your point, but in my (maybe short-sighted opinion) having a material that at least does something (in this case reading EventDisplat DL2 data and plotting the results against the "truth") is something usefull to give people a sense of what is the checkpoint from where to start "sprinting" .

HealthyPear: I think that with this PR and a very simple integration test we could be in a good position to go crazy and start more focused jobs

pyirf/scripts/make_DL3.py
HealthyPear: DONE

pyirf/scripts/make_DL3.py
MaxNoe: > it does everything

It creates DL3 event lists? For which of the input files? How is it determined which events are used for IRF calculation and which are used for the event list?

HealthyPear: The DL3 list is made by the DL2 data that pass the best cut-offs and that is used to create the IRFs

It creates one file per particle type with the suffix "_processed" in HDF5 format

pyirf/scripts/make_DL3.py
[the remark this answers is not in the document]
HealthyPear: here the "r68" option should correspond to the EventDisplay case, while "opti" should correspond to the CTA-MARS case am I right? (even if with different values)

HealthyPear:
OK, I have one remaining problem which came out with the docs after the recent updates from master.

Right now the versioning is read from `pyirf.version` into `docs/conf.py`
Before this was not the case, because in _read the docs_, having a `requirements` file for the docs alone, we didn't need to install _pyirf_ itself.

Now instead we need to (unless there is another way to read the version without importing the package?) and the build crashes when c tries to install the _pyirf_.

probably it is better to add a config file and tell read the docs to use `pip`?
In _protopipe_ I tried this and it seems `pip` tries to upgrade all packages making the build crash, while `setuptools` works fine.

For the time being a solution could also be to hardcode the versioning like "0.1.0-dev" and build the docs without requiring to install the package itself.

HealthyPear:
one of the recent PR added the reading of versioning from pyirf.version to setup.py like so

```python
with open("pyirf/version.py") as f:
    __version__ = re.search("^__version__ = '(.*)'$", f.read()).group(1)
```

but `re.search("^__version__ = '(.*)'$", f.read())` gives None.
Is it a bug in the regex?

Also, the current versioning is still 0.1.0-alpha, maybe we should update it to 0.1.0-dev or something similar?

MaxNoe:
It works for me on the current master. Also please see https://www.python.org/dev/peps/pep-0440/#version-scheme for valid versions. The `-` is not valid.

Also I don't think it matters right now, we are nowhere near an actual release, so alpha or dev or whatever does not matter.

pyirf/io/io.py
```diff
@@ -49,40 +87,179 @@ def read_simu_info_merged_hdf5(filename):
...
+        columns[key] = config["column_definition"][key]
+
+    if debug:
+        print("MAPPING TO GADF....")
+        print(columns)
+
+    return columns
+
+
+def read_FITS(config=None, infile=None, debug=False):
```
MaxNoe, line 173: As discussed elsewhere, there simply is no DL2 GADF data format.

I cannot stress this enough: we need to define our internal data format and then people can use the library for their data by putting it into those data structures.

pyirf/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.1.0-alpha'
\ No newline at end of file
+__version__ = "0.1.0-dev"
```
MaxNoe, line 1: - not allowed in python versions, see https://www.python.org/dev/peps/pep-0440/#public-version-identifiers

pyirf/io/io.py
HealthyPear: ouch , OK, I didn't get this problem fully before...

I guess that then this function is still doing it's job somehow (clearly, since the script works), but the context is wrong.

We can check and define the internal data format in a subsequent PR, but for this PR I would suggest then 

- to change function name (something like internal_dataformat_mapper) and its docstring to something that reflects this change in scope
- perhaps remove the check for the exitence of columns which then shouldn't be at all in DL2 simulated data

pyirf/version.py
HealthyPear: should this version be something like 0.1aN.devM ?
It is not clear to me what N and M are, though

pyirf/io/io.py
HealthyPear: also, I think I'd like to change `read_FITS` into `read_EventDisplay`, since the FITS HDUs structur is specific to that pipeline

HealthyPear: and in the config file, for the reasons you explained we should then remove only TIME, RA and DEC, correct?

pyirf/version.py
MaxNoe: These are counters, you can have multiple alpha versions, like 0.1.0a1, 0.1.0a2, but like I said, just leave it at 0.1.0 for now, it does not matter.

HealthyPear: OK

HealthyPear: Ah, side-question: I had problems building the docs on `readthedocs` when reading the version from `pyirf.version` because the installation of the package would fail.
I would like to fix this later, so for the moment I hard-coded the version in the docs to be the same as in version (I know, it's not a good thing...).

I was thinking though, can I copy-paste the lines used also in setup.py to open the file instead of importing it?
That way we still wouldn't need to install `pyirf` to build the docs and we can use the requirements file.

MaxNoe: I think the only problem there was that you need to move the `sys.path.insert` line before importing pyirf

pyirf/io/io.py
MaxNoe: Neither GLON/GLAT nor TIME/RA/DEC will make sense for any of the current simulated events.

pyirf/scripts/make_DL3.py
HealthyPear: I have opened issue #28 about the Pandas vs astropy tables discussions, so I leabe those comments for later reference and resolve this

pyirf/scripts/make_DL3.py
HealthyPear: removed incorrect comments and modified mapper from user to GADF (which was wrong) to an internal data format protoype

HealthyPear:
> It works for me on the current master. Also please see https://www.python.org/dev/peps/pep-0440/#version-scheme for valid versions. The `-` is not valid.
> 
> Also I don't think it matters right now, we are nowhere near an actual release, so alpha or dev or whatever does not matter.

Just to update: I fixed the versioning problem and left 0.1.0.

A good thing to do in order not to lose this working version for the coding sprint would be to tag the master branch after merging the current 2 PRs like 0.1.0-coding_sprint or something like this.

In this way, we will have for sure a usable tool (even if limited) and people can start playing with it, highlighting bugs or missing features while the longer work on refactoring takes place.

pyirf/scripts/make_DL3.py
vuillaut: I don't think that's a DL3 event list, it works only with simulated data and is only a by-product of the IRF production.
As a more general comment, it's not clear to me if py**irf** should produce the event lists. To be discussed elsewhere than here.

pyirf/scripts/make_DL3.py
vuillaut: Yes, the lstchain related script was a temporary thing to be able to produce the LST IRFs. It is bound to disappear.

pyirf/scripts/make_DL3.py
HealthyPear: It is not a DL3 list in the strictest sense (that would be "real" DL2 data to which the best cuts from the IRFs from pyirf are are applied)

in the case of pure simulated data pipeline (like the case we are approaching now) this data is the DL2 simulated data which passes the best cuts and therefore is used to create the IRFs

so, no I don't think we should create the DL3 event lists in this case (it is not clear to me when one would use those events in the end, am I missing something?)

should I rewrite the public docstring more specifically?

pyirf/scripts/make_DL3.py
HealthyPear: perhaps we should open an issue to merge the necessary information from that script into this one so (for the time being, minus future refactorings) we can also use LSTchain data and not only EventDisplay

docs/resources/index.rst
```diff
@@ -0,0 +1,147 @@
...
+   electron:
+    n_events_per_file: 450000000  #  10**5 * 20
+    n_files: 1
+    e_min: 0.005
+    e_max: 5
+    gen_radius: 1000
+    diff_cone: 1
+    gen_gamma: 2
+    offset_cut: 1.
+
```
vuillaut, line 91: I have to say I really don't like this config file.
But I will make a proposal in a separate PR.

HealthyPear: As with the rest, it is recycled material from protopipe, so we can change it later of course.
Yes, best to open a dedicated issue/PR for that.

MaxNoe: This goes in the same direction as other comments I made, but I actually think, we do not need a configuration system here.

We should aim for a  more library like use and expose everything via function / class parameters, not some configuration file or even system.

HealthyPear:
Ok, so we seem to have a problem with Travis, I guess due to their recent migration from .org to .com...
I tried everything, but the [PR on travis](https://travis-ci.com/github/cta-observatory/pyirf) is stuck at an older commit (related to d602102 from 2 days ago).

Possible solutions:
- I disable this merging requirementonly for this PR and you rely on the [build from my fork](https://travis-ci.com/github/HealthyPear/pyrf), which works fine for some reason,
- I close and reopen the PR, hoping it will catch up

MaxNoe:
Let me try to trigger the build. There were some strange emails and it seems even to have deactivated travis for ctapipe in between.

docs/resources/index.rst
HealthyPear: Why not? there are a lot of things that are many configurable quantities

MaxNoe:
Can you try pushing here again?

pyirf/scripts/make_DL3.py
HealthyPear: this is now explained better in issue #30

docs/resources/index.rst
MaxNoe: Because that adds a burden for potential users to adapt the configuration system we choose here. We should just avoid that and provide functions that are nice to use.

scipy for example also as algorithms with many configuration options, but it does not force you to use "the scipy config system"

MaxNoe:
Ok, seems travis is running now

docs/resources/index.rst
HealthyPear: perhaps it is not clear to me the difference between the two, we can certainly discuss this elsewhere

as long as the user can configure the optimization cuts and how IRFs production are produced it's fine

HealthyPear:
> Ok, seems travis is running now

what did you do to make it run again?
I also triggered the build manually and pushed a test commit, but it was just restarting from the older commit....

docs/resources/index.rst
MaxNoe: The difference is basically if the functions expect some global `config` object like it is now, or just the few options relevant to that function.

MaxNoe:
I did nothing.

MaxNoe (approved):
I would say let's merge this.

I want to stress that I'm not really happy with what is in here! I would not approve this PR normally.
it is just to large and does not really improve things. 

It mixes different things like refactoring, new code and just basic setup of CI / docs.
This should not be the case for any PR.

PR #25, opened 2020-09-10
Remove unused code and replace exec in setup.py

MaxNoe:
Just removed unused variables and imports (runnning `pyflakes` now has no complaints)

### DIFF
--- a/pyirf/io/io.py
+++ b/pyirf/io/io.py
@@ -3,8 +3,7 @@ import numpy as np
 from ctapipe.io import HDF5TableReader
 from ctapipe.io.containers import MCHeaderContainer
 import yaml
-import pkg_resources
-import os
+
 
 def load_config(name):
     """Load YAML configuration file."""
@@ -31,6 +30,7 @@ def read_simu_info_hdf5(filename):
         mc = next(mcheader)
 
     return mc
+
 
 def read_simu_info_merged_hdf5(filename):
     """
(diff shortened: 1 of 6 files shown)


pyirf/io/io.py
```diff
@@ -86,13 +86,3 @@ def get_simu_info(filepath, particle_name, config={}):
     print(cfg)
 
     return config
-
-
-
-# def get_resource(resource_name):
```
HealthyPear, line 92: This was added by @vuillaut , but I didn't see it used anywhere else is it stale code?

pyirf/scripts/lst_performance.py
```diff
@@ -157,7 +156,6 @@ def main(args):
                 continue
 
             psf = np.percentile(data['offset'], radius)
-            psf_err = psf / np.sqrt(len(data))
```
HealthyPear, line 160: Indeed this is not used, but I think we should keep it somewhere (I don't remember, is it defined in OGADF?)

pyirf/scripts/make_performance.py
```diff
@@ -130,7 +130,6 @@ def main():
                 # sys.exit()
 
             psf = np.percentile(data['offset'], radius)
-            psf_err = psf / np.sqrt(len(data))
```
HealthyPear, line 133: This script is going to be obsolete after PR #12 (make_DL3 is basically an updated version), you can ignore it in the future

HealthyPear:
Also it appear that this branch needs to be updated with (I guess) the previous PR changes that are now in master

pyirf/io/io.py
MaxNoe: It was completely commented out, so by definition not used.

pyirf/scripts/lst_performance.py
MaxNoe: The formula is also incorrect for estimating an uncertainty for the psf.

HealthyPear (approved):

Issue #21, opened 2020-09-10
About science tools


HealthyPear:
I wanted to have some input about whether or not - and if yes, how much - we should use any science tool feature _**inside**_ _pyirf_.

When I say "use", I actually want to emphasize the difference between _supporting_ a library and actually using it consistently to perform tasks (meaning making it a dependency).

I think we all agree that we need to support science tools.
(_gammapy_ and _ctools_ are the only players for now, am I wrong?)
This means that the DL3 output of _pyirf_ should be read by both in a format on which everyone agrees (which I guess is almost the case, since I think they already did some compatibility study between themselves).

What I would like to understand and trigger a conversation is whether we really need to call e.g _gammapy_ classes to compute (internally) and plot (internally or externally).
I think that performing these operations independently by any specific library would mean being less biased and / or dependent on upstream changes.

I this is not possible/advisable, we could also think of supporting multiple sources by calling their respective tools in a configurable way (and this would provide automatically a compatibility study from our side).

This discussion could be part of issues #10 and #17.

MaxNoe:
Supporting the science tools is *easy*: we will have a well-defined format for the IRFS that every tool needs to implement. 

pyIRF needs to emit this format, the science tools need to be able to ingest said format.

So this is not really a question I think.


On the question on wether a science tool should be a dependency of this library I would tend to answer **no**, we are one level below in the stack.

The question would be what features that a science tool provides would be needed / helpful here and weigh the advantage of just using that instead of (re-)implementing that feature vs. the inverted dependency situation.

MaxNoe:
So what features of gammapy do we currently use?

I think these are things like:

* spectral weighting
* statistical methods like calculating li & ma
* optimization / root finding

bregeon:
I agree Max on this point... may be just adding the comment again that if we want to rely on some I/O and/or plotting libraries then I see no good reason no to rely on `ctapipe` if needed.

HealthyPear:
> Supporting the science tools is _easy_: we will have a well-defined format for the IRFS that every tool needs to implement.
> 
> pyIRF needs to emit this format, the science tools need to be able to ingest said format.
> 
> So this is not really a question I think.

Exactly, I agree.

>
> On the question on wether a science tool should be a dependency of this library I would tend to answer **no**, we are one level below in the stack.
> 

Agree also on this

[...]

PR #36, opened 2020-09-18
Start refactoring pyIRF

MaxNoe:
Implemented in the first commit:

* reading of EventDisplay DL2 fits files
* Calculating event weights
* Functions for binning events
* Li&Ma Significance

Definition of the internal data format, for now astropy.table.QTables
with several columns expected for DL2 event lists.

Co-authored-by: Michele Peresano <michele.peresano@cea.fr>
Co-authored-by: Lukas Nickel <lukas.nickel@tu-dortmund.de>

More to come


pyirf/statistics.py
```diff
@@ -0,0 +1,50 @@
...
+    alpha: float
+        Ratio between the on region and the off region size or obstime.
+
+    Returns
+    -------
+    s_lima: float or array
+        The calculated significance
+    '''
+
+    scalar = np.isscalar(n_on)
```
kosack, line 29: why not use gammapy.stats for this rather than implement it again? It's already a dependency for the I/O

LukasNickel: Where is it used right now?

Anyway we want to get rid of it if possible (apart of plotting)

MaxNoe: Because we wanted to get rid of gammapy as a dependency. We won't need it for the IO.

kosack: OK, if you are removing gammapy as a dependency then keep it here.  But I would cross check the implementation with the one there. 

This is needed for both the optimization and for the sensitivity computation (to deal with the  5 sigma minimum per bin requirement), so we do need an implementation.

kosack: > won't need it for the IO.

You will re-implement all of it and the plotting as well? Seems like duplication of effort

HealthyPear: I updated issue #21 about this, I think what I propose is a good compromise.

MaxNoe: What I/O functionality is provided by gammapy?

MaxNoe:
Short update here: 

With the current version of the script in `exaples/calculate_eventdisplay_irfs.py`, we calculate a sensitivity that looks correctly calculated but which is different from the event display one due to different cut optimization, especially theta cuts are very different as what EventDisplay calculates does not seem to be a simple, bin-wise 68% containment.

pyirf_sensitivity


At the moment, we do: 

* Calculate theta cut as 68 % containment radius in very fine reco energy bins with a soft gamma hadron cut applied
* Optimize the gh cut in the bins of the sensitivity

HealthyPear:
>At the moment, we do:
> 
>     * Calculate theta cut as 68 % containment radius in very fine reco energy bins with a soft gamma hadron cut applied
> 
>     * Optimize the gh cut in the bins of the sensitivity

As far as I understood we are then only missing the 3rd step, which is to re-calculate the bin-wise 68% containment angular cut after the g/h bin-wise optimization

MaxNoe:
I found an error in reading in the data, the true azimuth was also set to the reconstructed one.
With that fixed and the recalculation of the theta cut, it now looks like this:

pyirf_sensitivity
 The two sensitivities are from the theta cuts before the gh cut optimization and after.

However, this does not change much and is now close to the theta cut table as written in the event display root files:

theta2_cut

MaxNoe:
Checking the event numbers, background rates unweighted are now in the single digits, mostly just one event. So this would explain the fluctuations in the central energy regime.

docs/changelog.rst
```diff
@@ -3,11 +3,7 @@
 Changelog
 =========
 
-This is the changelog for *pyirf*.
-
-We use a one-line description of every pull request, roughly in chronological order.
-We don't list every pull request.
-Maintenance and cleanup changes are not of interest to users and are not listed here.
```
HealthyPear, line 10: what was wrong with this information? it is also used in the _gammapy_ changelog

MaxNoe: Would for simplicity just list every PR, it is more easily extracted from the log, you don't have to judge
which ones are important and which are not.


Also: I find it demeaning to say "this is not of interest to users".
Users should be able to decide themselves what is interesting to them.

MaxNoe:
This is finally finished. Just a few tests remain to be written next week, but most is covered.

Documentation also got a large update. I think this is a very good starting point for the coding sprint.

Plots of the comparison with EventDisplay:

I will open another PR for the pypi upload and then we should release v0.2.0



Results from the EventDisplay comparison notebook:

effective_area
energy_resolution
gh_cut
psf
sensitivity
theta_cut


Implemented here:
- [X] Calculation of event weights
- [X] Point-Like IRFs (Effective Area, Energy Migration, PSF)
    - PSF and EDISP should also already work for full enclosure, but this is not tested yet
- [X] Sensitivity calculatation
- [X] Calculate cuts as percentile of a value (used for theta cuts here)
- [X] Evaluate binned cuts
- [X] Optimize GH cut for best sensitivity

vuillaut:
This is huge guys!
Nice work 👏
Do you want to undraft it and expect some review? I can have a look tomorrow...

MaxNoe:
I don't want to ask anyone to work on the weekend, but Michele and I wanted to take another look tomorrow and then merge to have it on PyPI for the coding sprint.

HealthyPear (changes requested):

README.rst
```diff
@@ -1,23 +1,13 @@
-==================================================
-pyirf |travis| |codacy| |coverage| |documentation|
-==================================================
+=========================================
+pyirf |travis| |coverage| |documentation|
+=========================================
```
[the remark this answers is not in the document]
MaxNoe: It did not work anymore, it did not appear, I just removed it because it was broken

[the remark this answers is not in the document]
MaxNoe: We should change the documentation one to github pages as soon as we merged and the docs are pushed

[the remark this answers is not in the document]
HealthyPear: > It did not work anymore, it did not appear, I just removed it because it was broken

on the master's readme it seems to work (https://app.codacy.com/gh/cta-observatory/pyirf/dashboard)
also in this PR is running
I would keep it and see what happens (non a real problem if we fix it later...)

> We should change the documentation one to github pages as soon as we merged and the docs are pushed

Yes sorry I didn't think that we can just change the link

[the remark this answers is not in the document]
HealthyPear: also I removed readthedocs, it should not pop out in the next PRs

HealthyPear:
Regarding the notebook:
- I don't see the PSF plot from the docs; is it just me?
- better add a` plt.show()` for each plot to get rid of useless matplotlib output (probably this solves the missing PSF plot, if it's not a problem on my local setup)
- some plots e.g. PSF shown in specific energy bins,  background rate and maybe even angular resolution could look better with a logaithmic y-axis?
- I would add this as the first cell (before the main title) in order to strip the cpde input cells from the doc page and show only markdown and actual output: this way I believe it is more easy to get a glance of the results and not get lost scrolling code that ca be seen by opening the notebook locally

```python
# Remove input cells at runtime (nbsphinx)
import IPython.core.display as d
d.display_html('<script>jQuery(function() {if (jQuery("body.notebook_app").length == 0) { jQuery(".input_area").toggle(); jQuery(".prompt").toggle();}});</script>', raw=True)
```

MaxNoe:
> better add a plt.show() for each plot to get rid of useless matplotlib output (probably this solves the missing PSF plot, if it's not a problem on my local setup)

That's not what plt.show does. Just by accident. Notebooks don't need a `plt.show`. To get rid of the annoying notebook, just add a semicolon after the last command or a `None` as the last line.

MaxNoe:
>     I would add this as the first cell (before the main title) in order to strip the cpde input cells from the doc page and show only markdown and actual output: this way I believe it is more easy to get a glance of the results and not get lost scrolling code that ca be seen by opening the notebook locally


I would show the code as it also demonstrates how to access the data in the output file.
I don't like hiding code.

MaxNoe:
> some plots e.g. PSF shown in specific energy bins, background rate and maybe even angular resolution could look better with a logaithmic y-axis?

I don't like adding unnecessary log scaling. For example you cannot judge the the correct normalization of the different bins for the psf. Angular resolution and background rate might be ok.

MaxNoe:
I added log scale and None as last line. I can see the PSF plot.

HealthyPear:
> I added log scale and None as last line. I can see the PSF plot.

yes me too now

HealthyPear:
I guess the changelog will get updated with the PR that uploads the package to Pypi?

MaxNoe:
Yes, let's do that in another PR preparing the release

README.rst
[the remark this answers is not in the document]
HealthyPear: > We should change the documentation one to github pages as soon as we merged and the docs are pushed

Actually we don't need to: the docs are built by the same CI pipeline, so if they fail the travis badge would already take it into account, no?

I would revert the change to have Codacy one (why did you think it didn't work?) and not have one for the docs

By the way, unless I am mistaken, we do not currently have an empty `gh-pages` branch were to deploy the docs when we merge - I think we should do that before merging

HealthyPear (changes requested):

MaxNoe:
> Actually we don't need to: the docs are built by the same CI pipeline, so if they fail the travis badge would already take it into account, no?

The badge should be for people to click on it to go to the documentation

HealthyPear:
> The badge should be for people to click on it to go to the documentation

Easier to click on the link in the "About" section of the main repo page no?
But if you can add one in the README is also ok

> It should be automatically created.

I see.

environment.yml
```diff
@@ -1,32 +1,27 @@
...
-  - sphinx-automodapi
   - sphinx_rtd_theme
   - pip
   - pip:
-    - rinohtype
-    - ctaplot
+      - rinohtype
+      - nbsphinx
+      - sphinx_automodapi
+      - uproot
```
vuillaut, line 27: move to uproot4?

MaxNoe: I talked with Jim Pivarski recently and uproot should become uproot4 soon. I would wait for that.

vuillaut: yes they plan to auto shift uproot to uproot4 this year so without a specific version the env might change without noticing

setup.cfg
```diff
@@ -2,18 +2,11 @@
 name = pyirf
 description = Python IACT IRF builder,
 url = https://github.com/cta-observatory/pyirf
-author = Julien Lefaucheur, Michele Peresano, Thomas Vuillaume
+author = Julien Lefaucheur, Michele Peresano, Thomas Vuillaume, Maximilian Nöthe
```
vuillaut, line 5: with this massive refactoring, I think you can remove the author list and use the contributors one from GitHub...

MaxNoe: This is in setup.cfg for the pypi metadata, not the docs. It should contain the names of the 
current maintainers, or  maybe even something like "CTA Consortium"

MaxNoe: I changed it to cta aswg

HealthyPear (dismissed):

HealthyPear:
we could add the codacy badga again though, for me it was working

MaxNoe:
If it works, great. Maybe codacy was just down for a while when I tried.

HealthyPear:
> If it works, great. Maybe codacy was just down for a while when I tried.

It should: the badge was pointing to [this url](https://app.codacy.com/gh/cta-observatory/pyirf/dashboard)- though, it's the master, the PR should be automatic I think

I will re-add it by copy-pasting the original RST command from their website just to be sure there wasn't a typo

HealthyPear (approved):

HealthyPear:
I just realized that we should probably add `cython` to `extras_require["tests"]`, otherwise _gammapy_ makes the installation crash for that section.

Can you confirm?

HealthyPear (approved):

HealthyPear (approved):

PR #46, opened 2020-09-27
Add PyPI deploy and metadata

MaxNoe:


HealthyPear (dismissed):

setup.cfg
```diff
@@ -1,12 +1,30 @@
 [metadata]
 name = pyirf
-description = Python IACT IRF builder,
+description = A python library to calculate IACT IRFs and Sensitivities
 url = https://github.com/cta-observatory/pyirf
 author = CTA Consortium, Analysis and Simulation Working Group
 author_email = thomas.vuillaume@lapp.in2p3.fr
```
vuillaut, line 6: I  can sure be the contact person if there is no other email but I think we may change that as well 😅 
aswg email?

HealthyPear: ups! I didn't see that...as fare as I know we are allowed only 1 email here, right?
In that case the aswg one could be a good choice

though, I don't think I would ever email the author for a problem, I would go to the repo...

MaxNoe: Is the aswg mailing list writable to for non-members? If yes, It would probably be a good choice.

HealthyPear: No idea...perhaps @bregeon knows

bregeon: For non-members, emails are forwarded to the list moderators, i.e. the list owners, i.e. the ASWG coordinators.
Eventually, this package will be owned and maintained by CTAO (or related to), so the contact person shall be either CTAO User Support or the "official" contact working as IkC for CTAO.

I am not sure what the best temporary solution is, we shall discuss with CTAO Computing if they have a preference.
Let see if we can come up with some agreement today.

MaxNoe: Let's just add me or michele for now until there is a more official one

bregeon: @MaxNoe yes, sounds good.

LukasNickel (approved):

setup.cfg
bregeon: For information, I have opened the issue with official instances, not only for pyIRF.

PR #52, opened 2020-09-28
Add sections about tests, coverage and building docs to docs

MaxNoe:


docs/contribute.rst
```diff
@@ -48,6 +48,73 @@ We use `Travis CI <https://travis-ci.com/github/cta-observatory/pyirf>`__ to
...
+.. code:: bash
+
+    make html
+
+Some changes require a full remake of the documentation, for that call
+
+.. code:: bash
+
+    make clean html
+
```
HealthyPear, line 100: I would add also that if one doesn't want to run all the notebook(s) the "clean" argument can be dropped

HealthyPear (changes requested):

docs/contribute.rst
MaxNoe: See the above comment about how normally to run

HealthyPear (dismissed):

HealthyPear (approved):

PR #58, opened 2020-09-28
Implement Background2d

### LINKED ISSUE #47: Implement Background rate IRF (symmetric case)
Description here: https://open-gamma-ray-astro.github.io/gamma-astro-data-formats/irfs/full_enclosure/bkg/index.html

MaxNoe:
fixes #47 

This is for now the radially symmetric version in the field of view

The tests for if gammpy supports reading is marked as xfail due to the axis order inconsistencies in gammapy, see 

https://github.com/gammapy/gammapy/issues/3025
https://github.com/gammapy/gammapy/issues/1221
https://github.com/gammapy/gammapy/issues/2067


pyirf/io/gadf.py
```diff
@@ -212,10 +212,73 @@ def create_energy_dispersion_hdu(
...
+)
+def create_background_2d_hdu(
+    background_2d,
+    reco_energy_bins,
+    fov_offset_bins,
+    extname="BACKGROUND",
+    **header_cards,
+):
+    """
+    Create a fits binary table HDU in GADF format for the PSF table.
```
HealthyPear, line 235: [...] for the **background** table.

pyirf/io/gadf.py
```diff
@@ -212,10 +212,73 @@ def create_energy_dispersion_hdu(
...
+    See the specification at
+    https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/psf/psf_table/index.html
```
HealthyPear, line 237: https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/bkg/index.html#bkg-2d

pyirf/io/gadf.py
```diff
@@ -212,10 +212,73 @@ def create_energy_dispersion_hdu(
...
+            "THETA_HI": u.Quantity(fov_offset_bins[1:], ndmin=2).to(u.deg),
+            # transpose as FITS uses opposite dimension order
+            "BKG": background_2d.T[np.newaxis, ...].to(GADF_BACKGROUND_UNIT),
+        }
+    )
+
+    # required header keywords
+    header = DEFAULT_HEADER.copy()
+    header["HDUCLAS1"] = "RESPONSE"
+    header["HDUCLAS2"] = "PSF"
```
HealthyPear, line 269: BKG

HealthyPear (changes requested):

pyirf/io/tests/test_gadf.py
```diff
@@ -85,3 +84,33 @@ def test_psf_table():
...
+    for point_like in [True, False]:
+        with tempfile.NamedTemporaryFile(suffix='.fits') as f:
+            hdu = create_background_2d_hdu(background, e_bins, fov_bins)
+
+            fits.HDUList([fits.PrimaryHDU(), hdu]).writeto(f.name)
+
+            # test reading with gammapy works
+            bg2d = Background2D.read(f.name, 'BACKGROUND')
+
+            assert u.allclose(background, bg2d.data.data, atol=1e-16 * u.Unit('TeV-1 s-1 sr-1'))
```
adonath, line 116: I think for the `Background2D` class in Gammapy the axis order is equivalent, so you can actually implement the test, no?

```python3
from astropy import units as u
from gammapy.irf import Background2D

energy = [0.1, 10, 1000] * u.TeV
offset = [0, 1, 2, 3] * u.deg
data = np.zeros((2, 3)) * u.Unit("s-1 MeV-1 sr-1")
data.value[1, 0] = 2
data.value[1, 1] = 4

bkg = Background2D(
    energy_lo=energy[:-1],
    energy_hi=energy[1:],
    offset_lo=offset[:-1],
    offset_hi=offset[1:],
    data=data,
)

print(bkg.data.data.shape)
print([ax.name for ax in bkg.data.axes])
```
Which prints:
```
(2, 3)
['energy', 'offset']
```

MaxNoe: The test here is meant to verify that the file / hdu can be read. So creating the in-memory version does not really help here.

MaxNoe:
Ok, the copy paste leftovers are fixed and I added a test for the background rate calculation.

pyirf/io/tests/test_gadf.py
MaxNoe: We opted to in-memory have the numpy convention for axis order like in the GADF doc page, which means we transpose when reading and writing from / to fits.

MaxNoe:
@HealthyPear could you review again? I think this is done.

HealthyPear (approved):
Just a reminder for future testing: wouldn't be easier to define a common toy event data set for all irf unit-tests? or 2, one for point-like and one for full-enclosure

MaxNoe:
@HealthyPear maybe, but here I chose the values so that the results can be easily checked by hand, I don't know if it is possible to come up with a single event set for all IRF tests with that property.

Issue #31, opened 2020-09-15
Plotting module


HealthyPear:
After discussing with @vuillaut, this is a proposal for a simple plotting module (aka `pyirf.plot` or `pyirf.vizualization`).

**Context:**

This module would provide a _simple_  and fast visualization suite to look into _pyirf_ products (even in-between, for testing and benchmarking) without medium/high knowledge of any science tools.

There is some overlap between this and _ctaplot_.
The latter could focus on sub-DL3 products, but it can still be useful to see e.g. the effective area before cut optimization (even though this can also be done by _pyirf_ since it applies the cuts on DL2 data after reading it).

**What we have:**
- in `pyirf.scripts.lst_performance` there are at least 1 function per IRF + a function to plot sensitivity using _ctaplot_ API components
- in PR #12 we have a notebook to plot all IRFs + sensitivity curve (artisanal plotting code made for fast testing)

One notebook should be sufficient, each plot saved separately or together, with a set of metadata.
Plotting could be done entirely by calling the specific _gammapy_ API (a few methods).

**What is missing:**
- a DL3 reader
- a configurator for plots (could be e.g. a dedicated config file or - better - part of the one already existing in PR #12 ) which then can be read via _papermill_ by the notebook
- the notebook could be interactive (binder, bokeh, ...)
- reference data like requirements (stored somewhere like for ctapipe+ctapipe-extra and called by the notebook)
- a function that plots a poster with IRFs and sensitivity curve EventDisplay-style (perhaps format A5, conference-ready)

bregeon:
Sounds like a very good plan, I am only wondering about the `gammapy` dependency.

For information, although that might be more useful for `ctaplot`?, we have currently two pieces of code related to plotting IRFs:
- https://gitlab.cta-observatory.org/cta-consortium/aswg/irfs-macros/sub-system-requirements
- https://gitlab.cta-observatory.org/cta-consortium/aswg/irfs-macros/public-performance-macros

HealthyPear:
> Sounds like a very good plan, I am only wondering about the `gammapy` dependency.

Yes, this is more or less the same discussion as for _ctapipe_.
Personally, I am not a fan of re-inventing the wheel and from one side I would prefer if _pyirf_ performed its internal operations (core utilities  - configuration - plotting) in a simple independent way from ctapipe and science tools...
Though in the case of plotting the situation is probably more elastic, because
- the scope of pyirf is not of plotting its output (this can be done anyway from the science tools)
[...]

PR #59, opened 2020-09-28
adding first plotting angular res function

vuillaut:
For early feedback

I will implement the other functions if this is the way to go in your opinion.

related to #31

### DIFF
--- a/pyirf/plot.py
+++ b/pyirf/plot.py
@@ -1,0 +1,36 @@
+import matplotlib.pyplot as plt
+from astropy.io import fits
+
+
+def plot_angular_resolution(irf_filename, ax=None, **kwargs):
+    """
+    Plot angular resolution from an IRF file
+    Parameters
+    ----------
+    irf_filename
+    ax
+    kwargs
+    Returns
+    -------
+    """
+
+    ax = plt.gca() if ax is None else ax
+
+    with fits.open(irf_filename) as irf:
+        psf_hdu = irf['ANGULAR_RESOLUTION']
+        energy = psf_hdu.data['true_energy_center']
+        energy_lo = psf_hdu.data['true_energy_low']
+        energy_hi = psf_hdu.data['true_energy_high']
+        psf = psf_hdu.data['angular_resolution']
+
+    kwargs.setdefault('fmt', 'o')
+    kwargs.setdefault('ms', 3)
+
+    ax.errorbar(energy, psf,
+                xerr=((energy - energy_lo), (energy_hi - energy)),
+                **kwargs,
+                )
+    ax.set_xscale('log')
+    ax.set_ylim(0, ax.get_ylim()[1])
+    ax.grid(which='both')
+    return ax


HealthyPear:
In #31 I was suggesting to use _gammapy_ methods from IRF objects to make it simpler and not reinvent the wheel, butat the same time, I do not know how much more flexible they are in the end.

If we think writing our implementation is better, I am all for it.

I guess we should just make sure that the different implementations do not clash.
Perhaps you can apply this function to the output file from the test data we have currently and superimpose it with e.g.
- the _gammapy_ equivalent
- the one we are using in the example notebook
- the one from _ctaplot_, if the one you provide is not already that one

pyirf/plot.py
```diff
@@ -0,0 +1,36 @@
+import matplotlib.pyplot as plt
+from astropy.io import fits
+
+
+def plot_angular_resolution(irf_filename, ax=None, **kwargs):
```
MaxNoe, line 5: I would use our in memory objects as input, not a filename.

vuillaut: right!

MaxNoe: so give it `true_energy_bins` and `angular_resolution`.

vuillaut: so that would be basically what exists already in `ctaplot`
I have no problem copy/pasting code - just to know what direction to take

MaxNoe: If a function in ctaplot already exists, great. I would just use it then. Not duplicate the code.

I am not to keen to have a lot of plotting code in here.

vuillaut: Ok but we agree to have a plotting module anyway to aggregate the existing implementation (ctaplot or gammapy) directly here?

MaxNoe: We could add that yes. Or just add an example script using either of those.

Providing wrappers around gammapy or ctaplot in `pyirf/plot` seems a bit unncessary to me, but if other's think it would be good to have, I am not against it.

HealthyPear: My point of view summarized:
- we should have a plotting module for internal fast check of the results without needing to _know_ gammapy
- no need to re-invent the wheel, so we could either import functions from _ctaplot_ or from _gammapy_
- if we diverge from (widely used?) packages like _gammapy_, let's make sure that we provide the same result

HealthyPear:
> I will implement the other functions if this is the way to go in your opinion.
> 
> related to #31

Sure!

Later on, we can call directly these functions in the notebook(s).

HealthyPear:
I will close this for the time being, because it's open for quite some time and we add the necessary functions to the notebooks.

The plotting "issue" is still open though: the I/O system and lack of configuration makes it suitable to analyze DL2 data from at least other IACTs.
So if ctaplot is supposed to be also as general as that, I would say let's use that otherwise we should implement our own plotting module (or rely on gammapy, but I would avoid that).
Related issues are currently #31 and in some way #21.

Issue #48, opened 2020-09-28
Add function to transform from energy dispersion to energy migration


maxnoe:
Energy dispersion is a better format for storage, as it is less sparse, but for actual computations, one needs energy migration.

We should have a function that transforms the energy dispersion matrix (Bins true energy, reco_energy / true_energy) to migration matrix (bins true_energy, reco_energy).

This is needed to estimate sensitivity from IRFs

LukasNickel:
So the difference is just in the value of the migration bins? The documentation uses these two terms interchangeably btw:
https://cta-observatory.github.io/pyirf/irf/index.html

MaxNoe:
Yes, we should make a clear distinction.

The difference is the second axis of the IRF, yes. One is the relative change one is absolute reconstructed energy.

HealthyPear:
perhaps the associated modification to the docs can be done together with the addition of this function

LukasNickel:
Is there an official document, that we could link for that?

MaxNoe:
@vuillaut I don't think this has anything to do with full enclosure irfs. 
It's not needed for the computation of the IRFs and should be the same procedure for point-like as for full-enclosure IRFs.

MaxNoe:
> Is there an official document, that we could link for that?


I don't think so. But there should be code in gammapy that does this transformation.

Maybe @adonath can give you a pointer

HealthyPear:
This ambiguity is also found in the OGADF website: energy _dispersion_ is defined as the PDF of _energy migration_

vuillaut:
> @vuillaut I don't think this has anything to do with full enclosure irfs.
> It's not needed for the computation of the IRFs and should be the same procedure for point-like as for full-enclosure IRFs.

Don't we want to compute the sensitivity from IRFs at the end?
Ok not directly related.

HealthyPear:
> Don't we want to compute the sensitivity from IRFs at the end?

Indeed, but I believe this is only important for cut optimizations which are not best-sensitivity (in that case the sensitivity depends on the specific optimized IRF no?)

MaxNoe:
> Don't we want to compute the sensitivity from IRFs at the end?
Ok not directly related.

Yes, it's a TODO for the sensitivity calculation from IRFs

LukasNickel:
> Energy dispersion is a better format for storage, as it is less sparse

This is because the transformation introduces additional bins outside of the "dispersion range" for a given true energy, right?
[...]

Issue #49, opened 2020-09-28
Implement histogram resampling


maxnoe:
For many calculations we will need resampling the finely binned IRFs to coarser bins for actual computation.

We should implement (a) function(s) to do this.

A possible approach for this is interpolating the cumulative sum.

adonath:
We also use this method in Gammapy, but it's hard-coded in our IRF classes where we need it. However it's not much code to re-implement. Here is an example implementation that I did some time ago to illustrate the method: https://github.com/adonath/notebooks-public/blob/master/histogram_resampling.ipynb 

Using `scipy.interpolate.interp1d` or possibly `RegularGridInterpolator `this can be extended to handle ND data as well, depending on how many axis you would like to integrate. In general I think this method is the best compromise between performance and precision. Using a suitable scaling for the data e.g. log-log for energy axes or sqrt for the theta ** 2 axes the accuracy of the method can be improved further...

nbiederbeck:
I would like to tackle this. Is it "enough" to implement the function from the notebook Axel posted? 

Also could you assign me, I am not able to do this myself?

LukasNickel:
I have already done this for the migration matrix without much optimization.
https://github.com/cta-observatory/pyirf/pull/64

If we have a more general method, that code can be used there aswell.

HealthyPear:
Was this issue fixed after #76 ? or do we want to add more implementations?

maxnoe:
I think we can close this for now, not clear if more is needed.

PR #64, opened 2020-09-28
Energy migration

LukasNickel:
Allows one to calculate an energy migration matrix (e_true x e_reco) from an energy_dispersion matrix (e_true x e_reco/e_true).
See issue #48

- The calculation is using scipy.interpolate.interp1d multiple times. If anyone knows a better way, please comment.
- I put the function in the `irf.energy_dispersion` file. I could also see it in `benchmarks` as its derived from the real irf, but its more of a different representation than a reduction I think.
- The whole dispersion <-> migration discussion is very poorly documented throughout the web. I added a small hint to the documentation, but it would be preferable to have a stringent distinction across all associated projects like OGADF, gammapy, ...


pyirf/irf/energy_dispersion.py
```diff
@@ -152,12 +128,33 @@ def energy_dispersion_to_migration(
...
 
-        migration_matrix[idx, :, :] = _interp_dispersion(
+        # for details of the interpolation see
+        # https://github.com/gammapy/gammapy/blob/f12ae6acece0ff5a6f269ca4f409f38c0714e23a/gammapy/irf/energy_dispersion.py#L242
+
+        # interpolate the cumsum of the dispersion values along the migration axis
+        cumsum = np.cumsum(e_true_dispersion, axis=0)
+        with np.errstate(invalid="ignore"):
+            cumsum = np.nan_to_num(cumsum / cumsum[-1])
+        f = interpolate.interp1d(
```
MaxNoe, line 142: Using `np.interp` should do the same and is faster:

```
In [7]: %timeit np.interp(x, xp, yp)
3.47 µs ± 10.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [8]: %%timeit
   ...: f = interp1d(xp, yp, kind='linear')
   ...: f(x)
   ...: 
   ...: 
45.9 µs ± 215 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

LukasNickel: What values did you interpolate there?
For me `np.interp` fails due to the additional `fov_offset` dimension.

pyirf/irf/energy_dispersion.py
```diff
@@ -73,3 +75,86 @@ def energy_dispersion(
...
+    """
+
+    migration_matrix = np.zeros((
+        len(new_true_energy_edges)-1,
+        len(new_reco_energy_edges)-1,
+        dispersion_matrix.shape[2],
+    ))
+
+    true_energy_interpolation = interpolate.interp1d(
+        (disp_true_energy_edges[1:] + disp_true_energy_edges[:-1])/2,
```
adonath, line 124: Usually energy bins are logarithmically spaced, so maybe take the geometric mean here? If not just ignore my comment...

pyirf/irf/energy_dispersion.py
adonath: The advantage of `interp1d` is the additional `axis` argument. You can interpolate along a given axis only and leave the other untouched. So you could replace the loop over true energy here...

pyirf/irf/energy_dispersion.py
```diff
@@ -73,3 +75,86 @@ def energy_dispersion(
...
+
+        # for details of the interpolation see
+        # https://github.com/gammapy/gammapy/blob/f12ae6acece0ff5a6f269ca4f409f38c0714e23a/gammapy/irf/energy_dispersion.py#L242
+
+        # interpolate the cumsum of the dispersion values along the migration axis
+        cumsum = np.cumsum(e_true_dispersion, axis=0)
+        with np.errstate(invalid="ignore"):
+            cumsum = np.nan_to_num(cumsum / cumsum[-1])
+        f = interpolate.interp1d(
+            (disp_migration_edges[1:] + disp_migration_edges[:-1])/2,
```
adonath, line 143: I think one should rather use there upper edges here, as the sum is defined up to the upper edge and no the center of the bin. Otherwise you shift the distribution by half an energy bin...

adonath (commented):
I have left a few technical inline comments.

LukasNickel:
Thanks for the comments.
We are already revisiting the interpolation with regard to
https://github.com/cta-observatory/pyirf/issues/49

pyirf/irf/energy_dispersion.py
LukasNickel: Yes, I was confused about the desired dimensions with respect to the `np.insert`.
Edges should work better, I agree

pyirf/irf/energy_dispersion.py
MaxNoe: Ah, I just used 1d. if scipy directly supports 2d, very good reason to use scipy

pyirf/irf/energy_dispersion.py
LukasNickel: About the true energy loop:
Isn't this needed, because the target binning changes for each true energy?

pyirf/irf/energy_dispersion.py
LukasNickel: About the true energy loop:
Isn't this needed, because the target binning changes for each true energy?

LukasNickel:
Superseded by #72

Issue #43, opened 2020-09-25
Use ogadf-schema to check if output is compliant to gadf


MaxNoe:
https://github.com/open-gamma-ray-astro/ogadf-schema

nbiederbeck:
Was this addressed with #63 ?

HealthyPear:
Technically yes, but some checks are also in #65.

The reason is that there is no explicit function for this, we just import that library and use its methods in unit-tests.

MaxNoe:
Yes, this was fixed in #63

Issue #24, opened 2020-09-10
Integration tests


HealthyPear:
This is probably the most urgent test to add for the coding sprint.

The test should check that the script (which after PR #12 is called make_DL3.py) is able to reach the end of the process and that the files are not empty.
This is basically what I do in `protopipe.scripts.tests` (actually, there the master branch is now outdated, but basically it does this).

It's not much but it would be already something useful to make automatize the development process.

Unfortunately, it is a test that requires test data to run the script on.
The Event Display datafiles I used are too heavy :

- 17M	electron_onSource.S.3HB9-FD_ID0.eff-0-CUT0.fits
- 122M	gamma_onSource.S.3HB9-FD_ID0.eff-0-CUT0.fits
- 123M	proton_onSource.S.3HB9-FD_ID0.eff-0-CUT0.fits

I don't know if there is a tool to extract a "sub"-FITS file from a bigger one or if we can have smaller datasets from EventDisplay.

MaxNoe:
Nitpick: This is not a unit test. A unit test tests a small **unit**.

Your are describing an integration test of the full system.

HealthyPear:
Ah thanks, I didn't know the name, so I put unit in quotes...

MaxNoe:
> I don't know if there is a tool to extract a "sub"-FITS file from a bigger one or if we can have smaller datasets from EventDisplay.

We can just create a smaller, sampled version of these files with a couple of thousand events.

This won't be able to calculate IRFs with small uncertainties in small bins but it could be used to test the chain with large bins.

MaxNoe:
Where are these files now and how can they be accessed? We could also provide a download script and and download them on travis using travis' credential system to provide a password if one is needed.

HealthyPear:
The URLs are linked in Issue #2 (bottom).

I put these URLs also in the usage/tutorial page in the new docs coming with PR #12 .

Regarding Travis doesn't each build have a storage limit? Do you mean the test files?

MaxNoe:
> Regarding Travis doesn't each build have a storage limit? Do you mean the test files?

Yes, but its several tens of GB: https://docs.travis-ci.com/user/reference/overview/

HealthyPear:
Ah ok then probably it's not a problem

Also the computing time right now it is relatively fast

MaxNoe:
> Also the computing time right now it is relatively fast

[...]

PR #65, opened 2020-09-28
Add integration test

HealthyPear:
This PR adds an integration test that runs **any** example named similarly to the current one (i.e. `calculate_*.py`).

It is part of issue #24.

Travis now runs the example from this test which:
- check that the output file exists and is not-empty,
- uses the ogadf-schema utilities to check that the HDUs are in the correct format (issue #43)

TODO:

- [ ] update documentation accordingly


pyirf/io/gadf.py
```diff
@@ -17,7 +17,9 @@
 
 DEFAULT_HEADER = Header()
 DEFAULT_HEADER["CREATOR"] = f"pyirf v{__version__}"
-DEFAULT_HEADER["HDUDOC"] = "https://github.com/open-gamma-ray-astro/gamma-astro-data-formats"
+DEFAULT_HEADER[
```
MaxNoe, line 20: Is this a black change? I don't really like it. It is less readable than before I think

pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
+import glob
```
MaxNoe, line 1: I'd like it better if we create a subdirectory `integration_tests` and this would be called `test_eventdisplay_example`.

pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
...
+import subprocess
+import logging
+import os
+from astropy.io import fits
+from ogadf_schema.irfs import AEFF_2D, EDISP_2D, PSF_TABLE, BKG_2D, RAD_MAX
+
+
+def test_integration_test(caplog):
+
+    caplog.set_level(logging.INFO)
```
MaxNoe, line 11: you can be specific about the logger to use. It should be `fits_schema` here.
Also, a minimum level of `WARNING` seems more appropriate. `fits_schema` uses `logging.WARNING` for unknown stuff, e.g. unexpected header keywords and `logging.ERROR` for violations of the standard.

pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
...
+
+    ROOT_DIR = os.path.dirname(os.path.abspath("setup.py"))
```
MaxNoe, line 13: This is not helpful as it will also only work in the root. It is basically the same as using `os.getcwd` since you use a relative path for `setup.py`. `os.abspath` will not look for a `setup.py`, it will just add the current working directory in front of the given path.

You can access the absolute path of the current file using the global variable `__file__` and move from there.

E.g. `ROOT_DIR = os.path.join(__file__, '../..')`

pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for example_script in glob.glob(f"{ROOT_DIR}/examples/calculate_*.py"):
```
MaxNoe, line 15: I don't like having this globbing here. We should have a single test function for each integration tests.

This makes it also clearer which function was executed.

pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # run script and check that it doesn't crash
+        subprocess.check_output(f"python {example_script}", shell=True)
```
MaxNoe, line 18: Why `check_output` and not run? Why the unsafe `shell = True`. Also: you should give the arguments as a list of strings, so they can be properly escaped: `sp.run([sys.executable, example_script]) would also make sure, that the file is run with the same python interpreter as the test script.

pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # check that the output file exists and it's not empty
+        outname = f"{ROOT_DIR}/pyirf_eventdisplay.fits.gz"
```
MaxNoe, line 21: use `os.path.join` for joining paths, or even better use `pathlib.Path`

MaxNoe: Also: here your globbing approach fails. other scripts probably won't have an output file name `pyirf_eventdisplay.fits.gz`

MaxNoe (changes requested):
See my inline comments

pyirf/io/gadf.py
MaxNoe: You can add a `# fmt: off` comment at the end of the line to keep black from interfering here.

setup.py
```diff
@@ -13,7 +13,16 @@
         "nbsphinx",
         "uproot~=3.0",
     ],
+<<<<<<< HEAD
```
MaxNoe: You need to resolve this merge conflict

pyirf/tests/test_integration.py
HealthyPear: I tried `sp.run([sys.executable, example_script])` and it was not working...

`check_output` is basically `run` but with a further check on the return code so it should be better than just running

`shell=True` could be removed since this script doesn't take any args it is sufficient to use a simple string like it is explained [here](https://docs.python.org/3/library/subprocess.html)

setup.py
HealthyPear: should be already solved

pyirf/tests/test_integration.py
HealthyPear: should we really need more than one integration test?
I thought we decided that there is no need for other example scripts.

perhaps we could have max 2 integration tests: one for point-like and one for full-enclosure?

HealthyPear: in that case I think 1 test_integration.py would suffice with such 2 tests inside

MaxNoe: Yes, in the end we should have integration tests for full enclosure and point like irfs and sensitivity.

But we can come up with more integration tests, e.g. if we implement different cut optimization schemes

HealthyPear: Wouldn't this go the current API-based approach we decided?
We shouldn't test any specific schemes, only check that given one example scheme (which we already have) we are able to produce correctly the required output.
Otherwise, we should have to take into account all the possible schemes, which is impossible.

HealthyPear:
I will reopen a sister PR, since this came from my fork and that's now failing due to the protected data test download.

PR #67, opened 2020-09-29
Add method integrating sim. events in FOV bins

MaxNoe:
Fixes #51

PR #68, opened 2020-09-29
Effective area 2d

MaxNoe:
To be merged after #67

### DIFF
--- a/pyirf/irf/effective_area.py
+++ b/pyirf/irf/effective_area.py
@@ -5,8 +5,8 @@ from ..binning import create_histogram_table
 
 __all__ = [
     "effective_area",
-    "effective_area_energy",
-    "effective_area_energy_fov",
+    "effective_area_per_energy",
+    "effective_area_per_energy_and_fov",
 ]
 
 
@@ -27,7 +27,7 @@ def effective_area(n_selected, n_simulated, area):
     return (n_selected / n_simulated) * area
 
 
-def effective_area_energy(selected_events, simulation_info, true_energy_bins):
+def effective_area_per_energy(selected_events, simulation_info, true_energy_bins):
     """
     Calculate effective area in bins of true energy.
 
@@ -50,7 +50,7 @@ def effective_area_energy(selected_events, simulation_info, true_energy_bins):
     return effective_area(hist_selected["n"], hist_simulated, area)
 
 
-def effective_area_energy_fov(
+def effective_area_per_energy_and_fov(
     selected_events, simulation_info, true_energy_bins, fov_offset_bins
 ):
     """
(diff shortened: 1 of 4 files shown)


MaxNoe:
No real tests yet for fully enclosure irfs. For that we need the diffuses gammas (or calculate a background effective area on the protons)

vuillaut:
> No real tests yet for fully enclosure irfs. For that we need the diffuses gammas (or calculate a background effective area on the protons)

I'd suggest to go on with protons, it really makes no difference for unit tests.

MaxNoe:
The unit test is already there. I was talking about an integration test where we could compare to some reference effective area.

HealthyPear:
Why we are not using anymore the _point-like_ vs _full-enclosure_ nomenclature?
This would comply also with the GADF naming convention.

MaxNoe:
>  Why we are not using anymore the point-like vs full-enclosure nomenclature?
This would comply also with the GADF naming convention.

We are using it. It just does not apply to these functions as both these functions can be used for point-like and fully enclosure psf depending on whether a theta cut has been applied to the input data or not.

pyirf/irf/effective_area.py
```diff
@@ -26,10 +27,9 @@ def effective_area(n_selected, n_simulated, area):
     return (n_selected / n_simulated) * area
 
 
-def point_like_effective_area(selected_events, simulation_info, true_energy_bins):
+def effective_area_energy(selected_events, simulation_info, true_energy_bins):
```
vuillaut, line 30: maybe `*_per_energy` would be more explicit?

pyirf/irf/effective_area.py
```diff
@@ -48,3 +48,38 @@ def point_like_effective_area(selected_events, simulation_info, true_energy_bins
...
+    """
+    area = np.pi * simulation_info.max_impact ** 2
+
+    hist_simulated = simulation_info.calculate_n_showers_per_energy_and_fov(
+        true_energy_bins, fov_offset_bins
+    )
+
+    hist_selected, _, _ = np.histogram2d(
+        selected_events["true_energy"].to_value(u.TeV),
+        selected_events["source_fov_offset"].to_value(u.deg),
```
vuillaut, line 78: couldn't the histogram be computed with units?

MaxNoe: It could yes, but normally it's faster to strip away units if possible. Here it is easy, since we know what units to convert too:


```
In [1]: import astropy.units as u
In [2]: import numpy as np
In [3]: energies = np.random.normal(100, 10, 10000) * u.TeV
In [4]: offsets =  np.abs(np.random.normal(0, 0.1, 10000)) * u.deg
In [5]: energy_bins = np.array([80, 90, 100, 110, 120]) * u.TeV
In [6]: offset_bins = np.arange(0, 1.0, 0.1) * u.deg

In [7]: %%timeit
   ...: np.histogram2d(energies, offsets, bins=[energy_bins, offset_bins])

410 µs ± 8.02 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [8]: %%timeit
   ...: np.histogram2d(energies.to_value(u.TeV), offsets.to_value(u.deg), bins=[energy_bins.to_value(u.TeV), offset_bins.to_value(u.deg)])

379 µs ± 6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Admittedly, the difference is not large here

pyirf/irf/effective_area.py
MaxNoe: yes

HealthyPear (approved):

PR #72, opened 2020-09-29
Feature resample histogram

nbiederbeck:
supersedes #64 and adds #49

### DIFF
--- a/docs/irf/index.rst
+++ b/docs/irf/index.rst
@@ -19,7 +19,7 @@ The energy dispersion matrix, ratio of the reconstructed energy over the true en
 as a function of the true energy, is computed with the events passing the
 threshold of the best cutoff plus the angular cuts.
 
-The corresponding, sparse energy migration matrix can be build from the dispersion matrix.
+The corresponding energy migration matrix can be build from the dispersion matrix.
 
 
 Point Spread Function
--- a/docs/irf/index.rst
+++ b/docs/irf/index.rst
@@ -15,9 +15,11 @@ the angular cuts.
 Energy Dispersion Matrix
 ------------------------
 
-The migration matrix, ratio of the reconstructed energy over the true energy
+The energy dispersion matrix, ratio of the reconstructed energy over the true energy
 as a function of the true energy, is computed with the events passing the
 threshold of the best cutoff plus the angular cuts.
+
+The corresponding, sparse energy migration matrix can be build from the dispersion matrix.
 
 
 Point Spread Function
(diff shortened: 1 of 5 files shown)


nbiederbeck:
I am trying ..

pyirf/binning.py
```diff
@@ -128,3 +129,58 @@ def create_histogram_table(events, bins, key="reco_energy"):
             events[key], bins, weights=events["weight"]
         )
     return hist
+
+
+def resample_histogram(data, old_edges, new_edges, axis=0):
```
MaxNoe, line 134: maybe add a `1d`?

docs/irf/index.rst
```diff
@@ -15,10 +15,12 @@ the angular cuts.
 Energy Dispersion Matrix
 ------------------------
 
-The migration matrix, ratio of the reconstructed energy over the true energy
+The energy dispersion matrix, ratio of the reconstructed energy over the true energy
 as a function of the true energy, is computed with the events passing the
 threshold of the best cutoff plus the angular cuts.
 
+The corresponding, sparse energy migration matrix can be build from the dispersion matrix.
```
MaxNoe, line 22: Sparse is not part of the definion. it's more like an expected property of the migration matrix.

The real difference is in the second axis: dispersion as `(e_reco / e_true)` , migration has `e_reco`

MaxNoe (changes requested):
Please add a test and fix the behaviour for bins that include under/overflow bin.

When I tested this locally with edisp that had under/overflow bins, all migration bins were `np.nan`

nbiederbeck:
> Does this now mean it is required to have an overflow bin?

No. There is a test with and without overflow bins. `np.nan_to_num` does nothing if there are no over/underflow bins

> And shouldn't it be `old_edges[-1]`?

`old_edges[-1] == np.inf` (with overflow bins), so we chose the second last bin which is not inf

nbiederbeck:
Travis still fails while downloading the test data

MaxNoe:
Ah, dam, this is because PRs from forks don't have access to the credentials.

HealthyPear:
> Ah, dam, this is because PRs from forks don't have access to the credentials.

Pardon my ignorance, but why would it matter?
Isn't this password just a string? if one has it it downloads otherwise not
or there is some other linked with information with it?

MaxNoe:
The download needs a password. That password is stored using travis' credential system. 

Build from pull requests don't have access to these credentials, because they could just add something like `echo $PASSWORD` to the .travis.yml to get the password.

So, only builds from the repo itself have access to these credentials for security reasons.

nbiederbeck:
after pushing to the main repo I need to open another PR ?

MaxNoe:
Probably, yes. I don't think there is the possibility to change the branch of a PR

PR #77, opened 2020-09-30
Calculate optimized cut on only the events surviving gh separation

MaxNoe:


LukasNickel:
I suppose this is just how eventdisplay performs the calculation of the cut?

MaxNoe:
Yes, and we thought we did this before, but I just forgot the mask.

LukasNickel (approved):

PR #80, opened 2020-10-01
Estimate unique runs taking pointing pos into account

MaxNoe:
This fixes an error of two we had because we assumed that unique `obs_id`s are sufficient for identifying the number of runs.

It is not, as different simulation sets start over with the same run numbers.

We now use unique combinations of `obs_id, pointing_az, pointing_alt`, which should be sufficient at least for the data we currently have. The new data should have the correct histograms, so we won't need that information at all.

### DIFF
--- a/pyirf/io/eventdisplay.py
+++ b/pyirf/io/eventdisplay.py
@@ -1,10 +1,9 @@
-from astropy.table import QTable
+import logging
+
+from astropy.table import QTable, unique
 import astropy.units as u
 
 from ..simulations import SimulatedEventsInfo
-
-import logging
-import numpy as np
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


MaxNoe:
Effective area before:
effarea



and after:

effarea2

LukasNickel:
> It is not, as different simulation sets start over with the same run numbers.

That sounds super weird. Any idea why this is done?

LukasNickel (approved):

GernotMaier:
> > It is not, as different simulation sets start over with the same run numbers.
> 
> That sounds super weird. Any idea why this is done?

Run numbers are limited in CORSIKA: Limit is: 0 ≤ NRRUN ≤ 999999

I don't understand why this matters. We could have the same run number for all runs, as long as we can identify the files in another way.

MaxNoe:
If I remember correctly,. and it is the only thing that makes sense, the CTA data model defines the combination of `OBS_ID, EVENT_ID` to be the unique identifier of an individual event.

This is broken here.

GernotMaier:
You implicitly assume here that CORSIKA run number == OBS_ID. I don't think this is the case.

MaxNoe:
It is at least in this case. Otherwise we wouldn't have had this problem.

PR #83, opened 2020-10-01
Update docs

HealthyPear:
In this PR I add again a URL link lost during the rewriting about the details of EventDisplay DL2 files used by the example script.


HealthyPear:
I will close this PR since now with the password-protected data test download it is not possible to push from a user's fork.

PR #85, opened 2020-10-01
Improve comparison to EventDisplay

MaxNoe:
* Do not require at least 10 weighted bg events
* Use correct IRF for comparison (before 180d Az, now both merged)
* Calculate background rate in same binning (does not effect sensitivity, only the plot in the notebook)
* Fix requirement of signal events larger than background (signal has to be 5 % larger than background, not larger than 5% of the background)

New sensitivity with this PR:

pyirf_sensitivity


LukasNickel:
>     * Calculate background rate in same bin

Can you elaborate how this is different from before?

MaxNoe:
* Can you elaborate how this is different from before?

We use twice the number of bins (just important for the plot, not the sensitivity)

LukasNickel:
>     * Fix requirement of signal events larger than background (signal has to be 5 % larger than background, not larger than 5% of the background)

But isn't this wrong? I thought the requirement was that the excess is at least 5 times the assumed background uncertainty (of 1%). What we label "signal" here, consists solely of the gammas, right?

MaxNoe:
> But isn't this wrong? 

No, the gammas *are the excess*.

n_on = n_signal + alpha * n_off
n_off = n_background
n_excess = n_on - alpha * n_off = n_signal

LukasNickel:
> > But isn't this wrong?
> 
> No, the gammas _are the excess_.
> 
> n_on = n_signal + alpha * n_off
> n_off = n_background
> n_excess = n_on - alpha * n_off = n_signal

Yes, so its either
- `n_on > 1.05 n_off`
- `n_signal > 0.05  * alpha * n_background`

But we compare `n_signal` and `n_background` here, or not?

MaxNoe:
No, n_signal must be *5 percent larger*. Not larger than *5 percent*. signal larger than 5% background does not make any sense, since then, signal < background.

MaxNoe:
Ok, turns out it was correct before, sorry

LukasNickel (approved):

HealthyPear (dismissed):

HealthyPear (approved):

PR #93, opened 2020-10-02
Cut update

MaxNoe:
* Use the 40 % efficiency cut as done by event display
* Add `min_events` parameter to the cut function calculation that requires at least that number of events to calculate the cut.
* Add smoothing option
* Add support for columns without units in `calculate_percentile_cut`

### DIFF
--- a/examples/calculate_eventdisplay_irfs.py
+++ b/examples/calculate_eventdisplay_irfs.py
@@ -127,7 +127,8 @@ def main():
         gammas["reco_energy"][mask_theta_cuts],
         bins=theta_bins,
         min_value=0.05 * u.deg,
-        fill_value=np.nan * u.deg,
+        fill_value=0.32 * u.deg,
+        max_value=0.32 * u.deg,
         percentile=68,
     )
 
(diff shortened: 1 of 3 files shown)


LukasNickel:
How are the 40% motivated? I thought eventdisplay uses 80%?

examples/calculate_eventdisplay_irfs.py
```diff
@@ -109,6 +109,8 @@ def main():
     background = table.vstack(
         [particles["proton"]["events"], particles["electron"]["events"]]
     )
+
+    INITIAL_GH_CUT = np.quantile(gammas['gh_score'], (1 - INITIAL_GH_CUT_EFFICENCY))
```
HealthyPear, line 113: wasn't this supposed to be a simple fixed cut?

HealthyPear:
> How are the 40% motivated? I thought eventdisplay uses 80%?

That is the maximum efficiency allowed later, this is instead a first global cut

examples/calculate_eventdisplay_irfs.py
MaxNoe: Yes, it is a single cut. The global cut which has 40 % efficiency. This calculates a single cut with that efficiency.

HealthyPear (approved):

PR #96, opened 2020-10-02
Add integration test

HealthyPear:
This PR adds an integration test that runs the EventDisplay example.

It is part of issue #24.

Travis now runs the example from this test which:
- check that the output file exists and is not-empty,
- uses the _ogadf-schema_ utilities to check that the HDUs are in the correct format (issue #43)

**Note**
This test operates on the current test data which corresponds to the _on-axis_ case.
Once we'll set up the off-axis DL2 test data to download, we'll need a new script (even by copying the current one).
That script will require its own integration test, so we should have in the end and integration test for point-like IRFs and one for full-enclosure ones.


pyirf/tests/test_integration.py
```diff
@@ -0,0 +1,50 @@
...
+    outname = os.path.join(ROOT_DIR, "pyirf_eventdisplay.fits.gz")
+    assert os.path.isfile(outname) and os.path.getsize(outname)
+
+    # open FITS file
+    outfile = fits.open(outname)
+
+    # known errors
+    dimensionality_error = "Dimensionality of rows is 0, should be 1"
+
+    # check that each HDU respects the OGADF schema
```
MaxNoe, line 30: Most importantly, we could no check that the sensitivity compared to eventdisplay is now:

```
ratio = sens_pyirf / sens_ed
assert np.all((ratio > 0.5) & (ratio < 2))
``` 
```

HealthyPear: ah yes now we can add this here

MaxNoe (approved):

MaxNoe (approved):

MaxNoe (approved):

PR #99, opened 2020-10-05
Set version to 0.3.0, add changelog

maxnoe:


docs/changelog.rst
```diff
@@ -36,7 +36,121 @@ We use a one-line description of every pull request.
...
++++++++++++
+
+This release is the result of the IRF sprint week in September 2020.
+
+Many bug fixes and improvements were made to the code.
+In the central and upper energy range, ``pyirf`` now reproduces the ``EventDisplay`` sensitivity
+exactly, the lower energy bins still show some disagreement.
+
+The cut optimization seems not yet to be the same as EventDisplay's and will be further investigated.
+
```
HealthyPear, line 73: I think we should be a bit more specific about the context to avoid misunderstandings: we now reproduce those results because we have a specific example that tries to follow that specific pipeline approach.

HealthyPear (changes requested):

docs/changelog.rst
maxnoe: Yes, agreed.

HealthyPear (approved):

PR #101, opened 2020-10-07
adding URL to CTA IRF prod3b doc

vuillaut:


HealthyPear:
The build fails because unfortunately, it is not possible to use forks anymore.
The CI pipeline is now automatically downloading protected test data.
You will need to clone this PR from a branch of cta-obervatory/pyirf and close this.

An additional comment: what if we add a `docs/references.rst` or `docs/bibliography.rst` were to store/call all references and URLs?

maxnoe:
I pushed the branch to this repo here: #136

PR #102, opened 2020-10-12
Update astropy to 4.0.2 to fix fits 1d issue

maxnoe:


setup.py
```diff
@@ -29,7 +29,7 @@
     packages=find_packages(),
     include_package_data=True,
     install_requires=[
-        "astropy~=4.0",
+        "astropy~=4.0,>=4.0.2",
```
HealthyPear, line 32: why keeping `~=4.0` if 4.0.2 is the minimal version that solves this probem?

HealthyPear (changes requested):
We should also remove any related material from the code, e.g. in the integration test we have code checking the rows dimentionality in the HDUs that will become stale with these changes.

setup.py
maxnoe: Because `~=` means any 4.x version. So this means "any 4.x" version larger than `4.0.2`

maxnoe:
Ok, I removed the "ignored_errors" part in the integration test as these are now expected to pass and they do

HealthyPear (approved):

PR #107, opened 2020-10-26
Small update to installation instructions

HealthyPear:


docs/install.rst
```diff
@@ -3,40 +3,42 @@
...
+the ``setup.py``.
 
-``pyirf`` requires Python ≥3.6 and the packages as defined in the ``setup.py``.
 Core dependencies are
 
 * ``numpy``
 * ``astropy``
 * ``scipy``
 
-Dependencies for development, like unit tests and building the documentation
```
maxnoe, line 14: Why remove this?

docs/install.rst
```diff
@@ -3,40 +3,42 @@
...
 
-If you want to work on pyirf itself, clone the repository or your fork of
-the repository and install the local copy of pyirf in development mode.
+If you want to work on pyirf itself, clone the repository and install the local
+copy of pyirf in development mode.
+
+It is required that the python environment contains also ``cython`` before
+installing pyirf.
+
+The dependencies required to perform unit-testing and to build the documentation
```
maxnoe, line 37: Ah moved, sorry, I think this is fine!

docs/install.rst
```diff
@@ -3,40 +3,42 @@
...
-
 Installing for development
 --------------------------
 
-If you want to work on pyirf itself, clone the repository or your fork of
-the repository and install the local copy of pyirf in development mode.
+If you want to work on pyirf itself, clone the repository and install the local
+copy of pyirf in development mode.
+
+It is required that the python environment contains also ``cython`` before
```
maxnoe, line 34: Please add that this is only required temporarily because of an issue with gammapy and add numpy, which is also required at install time.

docs/install.rst
```diff
@@ -3,40 +3,42 @@
...
+installing pyirf.
+
+The dependencies required to perform unit-testing and to build the documentation
+are defined in ``extras`` under ``tests`` and ``docs`` respectively
 
-Make sure you add the ``tests`` and ``docs`` extra to also install the dependencies
-for unit tests and building the documentation.
-You can also simply install the ``all`` extra:
+There requirements can be enabled by installing the ``all`` extra:
```
maxnoe, line 40: There?

environment.yml
```diff
@@ -1,15 +1,15 @@
-# A conda environment with all useful package for ctapipe developers
-name: pyirf-dev
+name: pyirf
 
 channels:
   - default
 dependencies:
-  - python=3.7
+  - python>=3.6
```
maxnoe, line 6: NO!

An environment can only have a fixed version. Using 3.7 is fine.

docs/install.rst
HealthyPear: right I forgot

maxnoe (approved):

PR #110, opened 2020-11-06
Optimize cuts in efficiency steps with maximum efficiency of 80% for EventDisplay comparison

maxnoe:


docs/notebooks/comparison_with_EventDisplay.ipynb
```diff
@@ -352,6 +352,8 @@
     "ax_sens.legend()\n",
     "fig.tight_layout(h_pad=0)\n",
     "\n",
+    "\n",
+    "fig.savefig('/home/maxnoe/sensitivity_cuts_optimized.png', dpi=300)\n",
```
HealthyPear, line 356: I think this is a leftover

docs/notebooks/comparison_with_EventDisplay.ipynb
```diff
@@ -386,7 +388,10 @@
     "\n",
     "plt.errorbar(x, y, xerr=xerr, yerr=yerr, ls='', label=\"EventDisplay\")\n",
     "\n",
-    "for name in ('', '_NO_CUTS', '_ONLY_GH', '_ONLY_THETA'):\n",
+    "# uncomment the other strings to see effective areas\n",
+    "# for the different cut levels. Left out here for better\n",
+    "# visibility of the final effective areas.\n",
+    "for name in ('', ): #'_NO_CUTS', '_ONLY_GH', '_ONLY_THETA'):\n",
```
HealthyPear, line 394: maybe too picky as a comment, but as it is written, someone could do this
`for name in ('', ): '_NO_CUTS', '_ONLY_GH', '_ONLY_THETA'):`
which is wrong syntax.
Better put those keys in a list above the for cycle and comment/uncomment single 
entries

examples/calculate_eventdisplay_irfs.py
```diff
@@ -60,6 +60,8 @@
 
 # Radius to use for calculating bg rate
 MAX_BG_RADIUS = 1 * u.deg
+MAX_EFFICIENCY = 0.8
```
HealthyPear, line 63: maybe better specify which kind of efficiency, since we specify e.g. at line 64

pyirf/cut_optimization.py
```diff
@@ -17,38 +18,95 @@ def optimize_gh_cut(
...
     progress=True,
+    **kwargs
 ):
     """
-    Optimize the gh-score in every energy bin.
-    Theta Squared Cut  should already be applied on the input tables.
+    Optimize the gh-score cut in every energy bin of reconstructed energy
+    for best sensitivity.
+
+    This procedure is EventDisplay-like, since it only applies a
```
HealthyPear, line 33: Now, this I think puts us in a weird position.

Either we rename this function by adding an `_EventDisplay` suffix or it should be added as a section of the EventDisplay script example.

HealthyPear (changes requested):

docs/notebooks/comparison_with_EventDisplay.ipynb
maxnoe: yes, indeed, sorry

pyirf/cut_optimization.py
maxnoe: Why do you think that? I think only optimizing the gh cut is avalid choice.

We can add other procedures and name them accordingly. But why do you think that mentioning that this procedure of optimizing the cuts is EventDisplay like is weird?

HealthyPear: No, I meant that it is weird if we do _not_ specify that is specific to EventDisplay (in some way, e.g. specifying it from the function name) because as the docstring says now, it is not a general API function as we wanted them to be.

maxnoe: Why isn't it?

maxnoe: The function says what it does, it is nothing specific to event display. It is just similar to what event display does.

HealthyPear: I would then do 2 things:
- simplify the docstring (no need to mention event display if there is no reference of some kind)
- rename the function in something more coherent with the doctring like `apply_theta_cut_optimize_gh_cut` or `optimize_gh_cut_with_thetacut`

because there will be other ways to optimize the cuts (different combinations)

maxnoe: You have to apply a theta cut to optimize a point source sensitivity so this is nothing special. The special thing is that this only optimizes the gh cut and takes a precomputed theta cut.

I don't see why the docstring shouldn't mention that this is what EventDisplay does. It is a fact and it is also the reason why we implemented this function in the first place.

When doing the full optimization in theta / gh / multiplicity, I am also going to mention that this is how MARS does it.

HealthyPear (approved):

PR #113, opened 2020-11-12
Add reconstructed energy as an optional independent variable for current benchmarks

HealthyPear:
In conformity with the IRF document, the angular resolution should be a function of reconstructed energy and not of true energy.

The binning should be the same as the one used for the Background Rate IRF.

**Reference:** 
Description of CTA Instrument Response Functions (Production 3b Simulation) v1.3
Page 13, Figure 5


maxnoe (changes requested):
No.

Angular resolution as a benchmark has to be vs. true energy. This is a similar discussion as the one for effective area.

Angular resolution vs. estimated energy mixes to IRFs, which harms its usefulness for comparisons.

I am completely opposed to changing this. I might be open to having it optional if it is required for some comparisons, but the default *must* be calculating it in bins of true energy.

Please also refer to the new IRF document, that was prepared during the sprint week.

vuillaut:
While I agree on the reasoning behind defining it vs true energy, it means that the produced angular resolution is not comparable with current official CTA ones. Is that ok?
@maxnoe can you point us to the new IRF document please?

HealthyPear:
OK.

but even in this version (which seems to be the last), this benchmark is shown against reco energy and this information fills the same table in the IRF files produced by ED and that we read from the notebook (i.e. AngRes)

so for the sake of the comparison, we cannot use true energy (we were in fact not comparing the right thing before)

I will try to modify it to make it optional

maxnoe:
> it means that the produced angular resolution is not comparable with current official CTA ones. Is that ok?

If that is the case, I would add the **option** to calculate it vs. estimated energy. It must not be the default.

> @maxnoe can you point us to the new IRF document please?

https://forge.in2p3.fr/projects/instrument-response-functions/files

HealthyPear:
It's a bit annoying, though, because e.g. we want to see it in true energy for the reasons above, but at the same time, we want to have also the values against the reconstructed energy for the comparisons.

Would it be so bad to calculate both and store 2 additional columns so one doesn't have to run the DL2 to DL3 step twice?

maxnoe:
@HealthyPear any user can of course calculate both and store both in a fits file. Where is the problem?

HealthyPear:
> @HealthyPear any user can of course calculate both and store both in a fits file. Where is the problem?
> 
> ```
> bias_vs_etrue, resolution_vs_etrue = bias_resolution(..., energy_key='true_energy')
> bias_vs_eest, resolution_vs_eest = bias_resolution(energy_key='reco_energy')
> ```
> 
> ?

that is that calculation, I was referring to the writing

should the user just create 1 BinTableHDU for each energy type then?

maxnoe:
> should the user just create 1 BinTableHDU for each energy type then?

Yes, why not? Or two separate fits files. Or two groups in hdf5 files. or two csv files. We deliver the table, the users decide what to do with it.

maxnoe:
@HealthyPear could you add the same change to the other benchmarks? Only doing it for angular resolution seems a bit inconsistent

HealthyPear:
> @HealthyPear could you add the same change to the other benchmarks? Only doing it for angular resolution seems a bit inconsistent

Sure.
Also in the [Prod3b-IRF-description](https://gitlab.cta-observatory.org/cta-consortium/aswg/documentation/internal_reports/irfs-reports/prod3b-irf-description) energy resolution is reported as a function of reconstructed energy.

maxnoe (approved):

PR #121, opened 2020-12-07
Upgrade support for uproot 4.0.0

HealthyPear:
Just a negligible fix of an RST formatting I missed last time....

**EDIT:** during this we discovered that uproot made a new release that required less negligible changes, so the scope of the PR got superseded.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ extras_require = {
         "pytest-cov",
         "gammapy~=0.18",
         "ogadf-schema~=0.2.3",
-        "uproot",
+        "uproot~=4.0",
         "awkward1",
     ],
 }
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ extras_require = {
         "pytest-cov",
         "gammapy~=0.18",
         "ogadf-schema~=0.2.3",
-        "uproot4",
+        "uproot",
         "awkward1",
     ],
 }
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ extras_require = {
         "gammapy~=0.18",
         "ogadf-schema~=0.2.3",
         "uproot~=4.0",
-        "awkward1",
+        "awkward~=1.0",
     ],
 }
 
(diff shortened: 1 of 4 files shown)


HealthyPear:
@maxnoe weird: I only modified the README, but I get an error from the docs when executing the notebook.
It seems to be something related to uproot, but on my local branch I do not get this error...have you ever noticed it in previous PRs?

maxnoe:
https://github.com/scikit-hep/uproot4/releases/tag/4.0.0 was just released, so maybe the API changed?

HealthyPear:
Seems so.
Should we then "embrace" 4.0.0 event though they say it's a "non-release-candidate"?
If we agree on this, I will remove all `errors` key arguments from that method and simplify the imports, since now uproot means uproot4 by default

maxnoe:
> Should we then "embrace" 4.0.0 event though they say it's a "non-release-candidate"?

Non-release-candidate means it is the first *real* release.

So yes, we should just require `uproot~=4.0` and do `import uproot` instead of `import uproot4 as uproot`

HealthyPear:
Don't mind the last merge commit, I mistakenly pushed into an old branch...

setup.py
```diff
@@ -20,7 +20,7 @@
         "pytest-cov",
         "gammapy~=0.18",
         "ogadf-schema~=0.2.3",
-        "uproot4",
+        "uproot",
```
maxnoe, line 23: Please add the version requirement `~=4.0`

maxnoe (changes requested):

maxnoe (approved):

setup.py
```diff
@@ -20,7 +20,7 @@
         "pytest-cov",
         "gammapy~=0.18",
         "ogadf-schema~=0.2.3",
-        "uproot4",
+        "uproot~=4.0",
         "awkward1",
```
maxnoe, line 24: Sorry missed that. `awkward1` should now also be `awkward~=1.0`

maxnoe (changes requested):
awkward needs to be updated as well.

setup.py
HealthyPear: done

maxnoe (approved):

PR #125, opened 2021-03-01
Interpolate irfs

jsitarek:
first implementation of the IRF (effective area and migration matrix) interpolation


pyirf/interpolation.py
```diff
@@ -0,0 +1,270 @@
...
+import pandas as pd
+import astropy.units as u
+from astropy.io import fits
+from scipy.interpolate import griddata
+
+# definition from lstchain.io.io copied here to avoid dependency
+dl2_params_lstcam_key = 'dl2/event/telescope/parameters/LST_LSTCam'
+
+
+def read_mean_pars_data(data_file, pars):
```
maxnoe, line 14: What are we parsing here?

maxnoe, line 14: This is an IO function using pandas to read in some custom data format. It has no place in pyirf. 

I only added support for reading these EventDisplay fits files reluctantly because we needed these for the integration tests. 

This should also not be needed to test the actual interpolation code.

pyirf/interpolation.py
```diff
@@ -0,0 +1,270 @@
...
+    Returns
+    -------
+    aeff_interp: astropy.units.Quantity[area]
+        Interpolated Effective area array with shape (n_energy_bins, n_fov_offset_bins)
+    """
+
+    n_grid_point, n_fov_offset_bins, n_energy_bins = aeff_all.shape
+
+    # get rid of units
+    aeff_all = aeff_all.to('m2').value
```
maxnoe, line 74: `.to_value(u.m**2)`

pyirf/interpolation.py
```diff
@@ -0,0 +1,270 @@
...
+    # interpolation
+    aeff_interp = np.empty((n_energy_bins, n_fov_offset_bins))
+    for i_th in range(n_fov_offset_bins):
+        for i_en in range(n_energy_bins):
+            aeff_interp[i_en, i_th] = griddata(pars_all, aeff_all[:, i_th, i_en], interp_pars, method=method)
+
+    # exp it and set to zero too low values
+    aeff_interp = np.exp(aeff_interp)
+    aeff_interp[aeff_interp < min_effective_area * 1.1] = 0  # 1.1 to correct for numerical uncertainty and interpolation
+    return aeff_interp * u.Unit('m2')
```
maxnoe, line 90: `u.Quantity(aeff_interp, u.m**2, copy=False)`

pyirf/interpolation.py
jsitarek: pars is for parameters
do you want the name of the variable to be changed?

pyirf/interpolation.py
```diff
@@ -0,0 +1,270 @@
...
+            for i_mig in r_mig:
+                matrix_interp[i_en, i_mig, i_th] = griddata(pars_all, matrix_all[:, i_th, i_mig, i_en], interp_pars, method=method)
+
+    # now we need to renormalize along the migration axis
+    norm = np.sum(matrix_interp, axis=1, keepdims=True)
+    mig_norm = np.divide(matrix_interp, norm, out=np.zeros_like(matrix_interp), where=norm != 0)
+    return mig_norm
+
+
+def read_unit_from_HDUL(hdul, ext_name, field_name):
```
maxnoe, line 135: Please just use `astropy.table.Table` to read in the IRFs, they automatically handle units.

pyirf/interpolation.py
jsitarek: I can move it to io, but we need the function to be able to use this IRF interpolation

maxnoe: Who is "we" here? LST? Than this function should go into lstchain, not pyirf.

maxnoe: Please don't use unnecessary abbreviations.

kosack: This is also related to the IRF format - it should not be necessary to go to the DL2 event list to look up the "bin location: of an IRF in simulation phase space.  If this is just to get the parameters of the single IRF, we should just require that this is metadata that is stored in the IRF format, either as header keywords or as a bintable if it is more complex.  

 But I agree it doesn't belong here, but also that in your LST code, you shouldn't have to open the event list to get this info - just store the metadata when writing the IRF.

pyirf/interpolation.py
kosack: and metadata! So if the parameters of the bin are in the IRF fits file headers, you automatically get them in the `.meta` attribute of your table.

pyirf/interpolation.py
maxnoe: I think this is opening the observed event list to get the point that needs to be interpolated at.

maxnoe: Which is completely unnecessary to implement the interpolation, because you can just choose where you want to interpolate. No need to have an actual data run from which you calculate the values.

pyirf/interpolation.py
```diff
@@ -0,0 +1,270 @@
...
+
+    # remove zeros and log it
+    aeff_all[aeff_all < min_effective_area] = min_effective_area
+    aeff_all = np.log(aeff_all)
+
+    # interpolation
+    aeff_interp = np.empty((n_energy_bins, n_fov_offset_bins))
+    for i_th in range(n_fov_offset_bins):
+        for i_en in range(n_energy_bins):
+            aeff_interp[i_en, i_th] = griddata(pars_all, aeff_all[:, i_th, i_en], interp_pars, method=method)
```
maxnoe, line 85: These loops are not neeed, since you can directly write:

```
aeff_interp = griddata(pars_all, aeff_all, interp_pars, method=method)
```

if my test with a multidimensional target function was correct.

pyirf/interpolation.py
kosack: Ah, sorry I was talking about the bin coordinates of each IRF, but I see now that these are instead reading the parameters that you want to interpolate *to*.    I'll open a separate issue about storing the bin coordinate in the output files.

We have not yet finished defining the official CTA *Observation Configuration* data model. Once that is available, it will be included in all DL1 & DL2 files, and can just be read in.   For now, this function seems to be using very LST-specific information, and so should just be in lstchain.

jsitarek: changed to "parameters"

pyirf/interpolation.py
jsitarek: done, thanks

pyirf/interpolation.py
jsitarek: done, thanks

pyirf/interpolation.py
jsitarek: Hi @maxnoe, @kosack 
The whole point is that we want to interpolate the IRFs for a given run, so such a function is a part of the implementation. But I agree with you that it was LST-specific. Therefore I changed it now, removing the LST-specific key from it. Now it can be used with whatever data as long as they are in FITS format. Even while it is more "io" then "interpolation" I kept it for the moment in interpolation.py, because in io directory it does not really fit into gadf.py, nor in eventdisplay.py, and making a separate file for a single function is excessive in my view.

maxnoe: No. We explicitly decided that functionality like this will not be part of pyirf. 

pyirf should just provide a function that can interpolate IRFs, given a grid of IRFs and an interpolation point. Where this interpolation point comes from should not matter to pyirf and has to be part of the pipeline.

Changing this decision would not only affect the interpolation part but much more, so please don't introduce such features in this PR. It is not needed to implement the interpolation of IRFs.

jsitarek: can you point me to where is it documented that pyirf should not read from the data parameters to which the IRFs should be interpolated?
either way in the 7e3f197..82d3135 commit I remove the function from the bulk of pyirf code and moved it in the example file which shows how to use the interpolation functions

maxnoe: Specifically for interpolation, it is not documented because you only started to implement interpolation at all.

But the general design decisions are documented here:https://cta-observatory.github.io/pyirf/introduction.html

Especially the general decision to not implement reading of any specific DL2 event list format but to leave that to the pipelines using pyirf. 

The event display fits files are an exemption here because they are used for the integration test.

maxnoe: For the same reasons, we decided not to implement any kind of configuration system and just provide all necessary parameters as keyword arguments to the functions.

Like this, each user of pyirf can use their configuration system to call the pyirf functions with the correct arguments.

maxnoe: For interpolation, it is very simple. The pipeline calls the pyirf function to interpolate with the correct interpolation point and grid irf data. And this is all that should be implemented in this PR.

jsitarek: Honestly @maxnoe , I have troubles following your logic. You say "general decision to not implement reading of any specific DL2 event list format", while the document to which you point me says something opposite: "Most functions for calculating IRFs need DL2 event lists as input. " than you make an exception for the eventdisplay functions which are really special files (and you even need to translate the column names there), but you do not want a function that is directly connected with interpolation and it operates on any kind of DL2 fits file.

still, I do not want to argue over this function, as I wrote above I moved it already from the bulk of the pyirf code to the example script, but this again goes opposite to what you were telling me before to provide functions rather then macro 

I would just recommend you to specify more clearly what pyirf should and what should not do, describe it somewhere accessible to developers. This would safe time of both the developers and code reviewers.

maxnoe: @jsitarek Event lists as a astropy table with specific column names as opposed to event list *file formats*, that's a big difference.

We don't want to write parsing code for each possible event list format. So we take the already parsed event lists as an astropy table.

pyirf/interpolation.py
jsitarek: thanks for the tip on this @maxnoe, I implemented your solution and thus was able to remove this function

@kosack, as you say the meta data should be then read as well, but I'm not sure how useful those will be in reality. For the moment interpolation is in cos zd and Az, but the functions can be used for the whole zoo of interpolations in NSB, PSF parameters, etc. I do not think all of this will end up in the meta data, and even if so one would still need to relate the values stored in the meta data of IRFs with the values stored in the event lists of the data. Let's come back to this discussion once we have IRF files with meta data

pyirf/interpolation.py
jsitarek: @maxnoe but this kind of parsing is exactly what you were doing in eventdisplay exception, while the function that I implemented did not have any special parsing. The user specified what is the interpolation variable and where it is and the code was computing its average value that will be used in the interpolation. The function do not need to worry about the data format, user (or pipeline) that uses it has to specify it.

maxnoe: But it is completely unnecessary for implementing and testing this code. Why do you need to read in actual LST run data when you can just use any specific value for the tests / the example?

Why not simply write:

```
mean_altitude = 60 * u.deg
mean_azimuth = 20 * u.deg 
```

then all this additional code is unnecessary.

Please concentrate on what is actually necessery *here* to implement this interpolation.

From my point of view, the only thing needed is one function for each kind of IRF we want to interpolate that do the corresponding interpolation and tests for these functions. Nothing else.

jsitarek: Such a test (with a dummy shape of Aeff and dispersion matrix) I already implemented. 

I'm implementing the IRF implementation because it is needed (now for LST, but also for other parts of CTA), and if one wants to use it a function reading the information from the data, no matter if this is LST, SST, or whatever else will be needed as well. I though it would be useful and clean for the user to have this function together with the interpolation code.

HealthyPear: Hi all,

@jsitarek first of all thanks for your contribution!

I didn't have time to read everything from the issue and review all the code, but for what concerns this function I think we can solve this relatively easily.

To re-iterate what @maxnoe said: _pyirf_ has been built as an API made of simple functions that are supposed to be used in external scripts belonging to pipelines or other code.
Said this, it is true that we had to start somewhere and we implemented such a script for EventDiplay as an example with associated I/O function.

What is important to get from the I/O part of pyirf (might be good to review that part of the docs if necessary) is that what _pyirf_ "first" functions need to read is essentially the output as from `pyirf.io.eventdisplay`.

We do not want to add another function like that (which is only an example, and I am open to consider to change its description more as a "template" example).

What I propose (I guess as a mid-way solution between what you and Max discussed) is to change the name of this function from `read_mean_pars_data` to e.g. `get_mean_DL2_parameters` and make it "eat" the same data as output from `pyirf.io.eventdisplay`, so

- 1 `astropy.QTable` for DL2 events
- 1 `pyirf.simulations.SimulatedEventsInfo` instantiation

Then these 2 ingredients will be produced by `pyirf.io.eventdisplay`-like functions in `cta-lstchain` or `protopipe`.
What do you think @maxnoe ?

I apologize in advance if my description was somewhat shallow, I will delve into the related issue/PR ASAP.

maxnoe: > wants to use it a function reading the information from the data, no matter if this is LST, SST, or whatever else will be needed as well.

But you mix several things here. Reading values from a file (hdf5 using pandas, which is not used at all in pyirf!) and aggregating those values according to some function.

This are independent operations. And `pd.read_hdf` is by no means flexible. For example it will ignore unit information that is present in e.g. files created by the ctapipe table writer.

> Then these 2 ingredients will be produced by pyirf.io.eventdisplay-like functions in cta-lstchain or protopipe.

Which already exist!

pyirf/interpolation.py
```diff
@@ -77,16 +74,8 @@ def interpolate_dispersion_matrix(matrix_all, pars_all, interp_pars, method='lin
...
-    r_mig = range(n_migration_bins)
-    matrix_interp = np.empty((n_energy_bins, n_migration_bins, n_fov_offset_bins))
-    # TO DO this part can be optimized because nested for loops take quite some time
-    # but it is not a big problem because this has to be done only once per run
-    for i_th in r_th:
-        for i_en in r_en:
-            for i_mig in r_mig:
-                matrix_interp[i_en, i_mig, i_th] = griddata(pars_all, matrix_all[:, i_th, i_mig, i_en], interp_pars, method=method)
+    matrix_interp = griddata(pars_all, matrix_all, interp_pars, method=method)
+    matrix_interp = np.swapaxes(matrix_interp, 0, 2)
```
maxnoe, line 78: Nice!

maxnoe: Glad you found a way to get rid of these loops.

pyirf/interpolation.py
jsitarek: thx @maxnoe I implemented it this way (just with additional axis transposition that was needed) and also did the same with dispersion matrix

I did not realize that this will work because https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html says that values should have shape (n,), and here obviously it is not a simple vector. Nevertheless when I run it over the dummy test it works and provide the same output, and running it over the example makes it much faster (from 140s to 0.013s for 0.9M of interpolations !)

pyirf/interpolation.py
jsitarek: @HealthyPear the function is literaly a few lines of code, if you separate the reading of the data into another function what is left is only computing the average that can be done with one line thus making a function for this is unnecessary complication in my view

@maxnoe, the function has one task - compute the values to which you need to interpolate, and it does so by reading the data and averaging, I see no problem in this. pd.read_hdf is flexible in the sense that it can evaluate simple formula and mathematical operations done on the data in the table, where the information to be extracted is parsed from a string. This is very desirable, because we do not need to hardcode the variables over which the interpolation is to be performed, but they can be read in (as they are in the example macro) from a config file. The fact that it ignores the units is not a problem and actually makes things simpler, because since the user specifies in the config file where the variables are stored he obviously must know what is there and with what units. While pandas has not been used so far in pyirf, it is used in ctapipe, so effectively for the people using it, it would not cause any additional dependancy (and either way it is a pretty standard package) 

Let's try to converge on this, since this seems to be the only remaining issue of the review. As explained above, despite I still think the function would be useful in general pyirf code, I removed it from there already and only kept it in the example macro showing how to properly produce the interpolated IRFs. @maxnoe @HealthyPear please let me know if this solution is fine with you. If not the only thing that I can do is to move the code of the function directly where it is used.

maxnoe: I will try to review fully tomorrow. But I have fundamental issues with the current state here. We need to discuss the big picture a bit further before we can actually go into the detailed review of the code.

In short, please remove this IO function and don't add pandas. ctapipe does not use pandas. pyirf will not use pandas if we don't have a *very* strong reason to do that. And currently, I don't see it.

I will try to give a more complete feedback tomorrow on the big picture. I think we went too soon into actual code review before discussing the grand scheme of things.

kosack: Hi just to comment here. The idea is to decouple IO and pipeline-specific code and data formats from IRF algorithms.  As currently implemented, this sort of IO function is very pipeline-specific (at least until we have further standardized DL2 data), so doesn't belong in a general IRF library.   I agree it should go somewhere, though.  Once we have a standard DL2 format, it could be in ctapipe, for example.     The interpolation function itself however, is very useful to have in pyirf. 

The final goal is that you write an *IRF creator* and *IRF combiner*  command-line tools are implemented using the framework of the pipeline itself (so either ctapipe, once standardized, or for now in protopipe or lstchain), and those tools import `pyirf` as a library.  This avoids circular dependencies, and keeps PyIRF clean of pipeline-implementation-specific details.  

So what you are writing here is very useful, but should be split into two parts: the generic part is in pyIRF, and the pipeline-specific parts that have to access DL2 files go in ctapipe.tools, protopipe, or lstchain, depending on the use case. 

However, I would say that anything that accesses DL3 IRF files *generated by pyirf* can go in pyirf.io, since then it is self-consistent So the rule would be:
 * anything to do with reading back in the FITS files that PyIRF outputs in a generic way (not including any pipeine-specific parametres)  is fine in `pyirf.io` 
 * anything that needs pipeline-specific info or any data file other than DL3 FITS files should be in the pipeline code itself 

That is why there is no tool in pyirf to generate IRFs: it is a library of functions that lets you implement your own tool in your pipeline.

HealthyPear: > @HealthyPear the function is literaly a few lines of code, if you separate the reading of the data into another function what is left is only computing the average that can be done with one line thus making a function for this is unnecessary complication in my view

Precisely, that is exactly the point :)
It's still good to have 1 function to apply the averaging, it can be used by other interpolators or marginalizing algorithms that we might add in the future.

I would also add that this is a perfect analogous case to what I always did while developing for _protopipe_/_ctapipe_.
First I test new code in _protopipe_, then once it has been tested through an analysis I migrate the library-specific code to _ctapipe_/_pyirf_, thus filtering automatically all the data-specific part of the pipeline (in this case the DL2 format).

What has to enter in _pyirf_ needs to be in the format which we call "in-memory" which is basically the one output by the example IO function for _EventDisplay_ data. 

I would go even further and say that even the JSON file should not stay in _pyirf_, but the interpolation function (in general any marginalizing function) should read dictionaries with parameters as keys and values as Quantity arrays that come from a config file read by the pipeline (exactly like it happens for e.g. `ctapipe-stage1`)

> since the user specifies in the config file where the variables are stored he obviously must know what is there and with what units

If there is something I learned since I started coding on CTA code is that I should never trust the user (specifically me, since I am the only one testing my code!).
we should use quantity and units everywhere possible to limit mistakes, especially because CTA-code will be used by many more people in the future.

> I still think the function would be useful in general pyirf code, I removed it from there already and only kept it in the example macro

I think this too, even if it is one line that makes an average, but if it is used more than once or by more than 1 function ti makes sense to add it to the API of _pyirf_, but it needs to read/output in-memory objects.

jsitarek: @maxnoe I removed the function from the example macro, but pandas is still needed because QTables do not have the required functionality that I explained above
contrary to what you wrote ctapipe is using pandas:
https://github.com/cta-observatory/ctapipe/blob/70b3ab5ba5c606fc06493e25f24528f25aac0f3c/environment.yml#L25

@kosack as discussed above the rule that you mention is already violated for eventdisplay.py, where translation from event display naming to pyirf naming of variables is done. In the case of my (already removed) function, the function was made to be non-pipieline-specific, however the value of parameters  was (but this is normal and just shows that the function is universal). Concerning moving things to pyirf.io I could move read_fits_bins_lo_hi, read_irf_grid,  compare_irf_cuts to this directory, either adding them to gadf.py or making a new file like read_irf.py. Those three functions clearly qualify for your first case

maxnoe: @jsitarek pandas is needed to test ctapipe generated hdf5 files are *readable* by pandas. ctapipe does not use pandas as a dependency for actual ctapipe code. Arguing about this is besides the point.

Please accept that we don't want pandas as a dependency and in general that this functionality should not be part of this PR.

Let's concentrate first on getting the actual interpolation implementation, like I asked from the beginning.

That means what @HealthyPear wrote and I wrote above. The only thing that should be in this PR should be:

* Functions that take an array of multiple IRF components, the corresponding grid points and the interpolation point and return the interpolated results
* tests for these functions.

Can we start with this and leave everything else to the side for now please?

jsitarek: > > @HealthyPear the function is literaly a few lines of code, if you separate the reading of the data into another function what is left is only computing the average that can be done with one line thus making a function for this is unnecessary complication in my view
> 
> Precisely, that is exactly the point :)
the function would be np.average(..., axis=0), do you really want to make a wrapper for such a one line function?? 
the nice thing of the function in my view was being able to average from the file that user wants any formula that user wants 


> It's still good to have 1 function to apply the averaging, it can be used by other interpolators or marginalizing algorithms that we might add in the future.
> 
you use it only once and at the very beginning, when you determine to what you want to interpolate to

> I would also add that this is a perfect analogous case to what I always did while developing for _protopipe_/_ctapipe_.
> First I test new code in _protopipe_, then once it has been tested through an analysis I migrate the library-specific code to _ctapipe_/_pyirf_, thus filtering automatically all the data-specific part of the pipeline (in this case the DL2 format).
> 
I was explictely asked to implement this in pyirf and not in lstchain

> What has to enter in _pyirf_ needs to be in the format which we call "in-memory" which is basically the one output by the example IO function for _EventDisplay_ data.
> 
and this is exactly what happens here. the interpolation functions in interpolation.py are using such a format. Then there are a few IO functions that however deal with IRF DL3 format as Karl said and those can go into io directory

> I would go even further and say that even the JSON file should not stay in _pyirf_, but the interpolation function (in general any marginalizing function) should read dictionaries with parameters as keys and values as Quantity arrays that come from a config file read by the pipeline (exactly like it happens for e.g. `ctapipe-stage1`)
> 
the interpolation functions are taking parameters from arrays, json file is only used in the example showing how to run it

> > since the user specifies in the config file where the variables are stored he obviously must know what is there and with what units
> 
> If there is something I learned since I started coding on CTA code is that I should never trust the user (specifically me, since I am the only one testing my code!).
> we should use quantity and units everywhere possible to limit mistakes, especially because CTA-code will be used by many more people in the future.
> 
My understanding is that the general user will start with DL3 file, everything before (including this IRF interpolation) needs some level of "cherenkov expert"

maxnoe: >  I was explictely asked to implement this in pyirf and not in lstchain

No, you were asked to implement the actual interpolation code in pyirf. Not to implement everytrhing needed to interpolate LST IRFs into pyirf.

maxnoe: > My understanding is that the general user will start with DL3 file, everything before (including this IRF interpolation) needs some level of "cherenkov expert"

And this is justification for dealing with units manually instead of relying on astropy? come on.

maxnoe: > average from the file that user wants any formula that user wants

1. you can do that without pandas
2. you again mix file IO with computing stuff. These things need to be separated. One function to read event lists from a file, one function to compute arbitrary stuff on an arbitrary column of a table. And for pyirf, we decided that the first should in any case not go into pyirf and for the second I really don't see the need at this stage.

jsitarek: > @jsitarek pandas is needed to test ctapipe generated hdf5 files are _readable_ by pandas. ctapipe does not use pandas as a dependency for actual ctapipe code. Arguing about this is besides the point.

I think you are wrong about this Max, here you have an examples of ctapipe code that use pandas:
https://github.com/cta-observatory/ctapipe/blob/70b3ab5ba5c606fc06493e25f24528f25aac0f3c/ctapipe/plotting/charge_resolution.py
https://github.com/cta-observatory/ctapipe/blob/70b3ab5ba5c606fc06493e25f24528f25aac0f3c/ctapipe/tools/extract_charge_resolution.py

maxnoe: I just removed that code in a PR because of an issue raised by Karl the other week. And again, this is besides the point.

maxnoe: If ctapipe uses pandas or not is completely irrelevant to the discussion here.

jsitarek: > > I was explictely asked to implement this in pyirf and not in lstchain
> 
> No, you were asked to implement the actual interpolation code in pyirf. Not to implement everytrhing needed to interpolate LST IRFs into pyirf.

Quoting Abelardo:
"I think what we need (but please anyone more familiar with pyirf feel free to comment) is a tool in pyIRF to interpolate IRFs (Aeff, mig matrix, PSF?) calculated with MC for different directions, then for each real data run we have to check the distribution of directions, and obtain the best-fitting IRF using the interpolation tool. "
and this is precisely what I did, and yes I did this the way you say, i.e. separating all the interpolation functions, IRF i/o functions, and separating an example file that shows how to combine it all

maxnoe: Tool here was not a `ctapipe.core.Tool` but tool as in the common english word.

maxnoe: @jsitarek I am grateful that you are putting your time into this and the interpolation approach using `scipy.interpolate.griddata` is certainly much appreciated. 

But please, can we reduce this PR to just the functions that we have no discussed many times?

jsitarek: > > My understanding is that the general user will start with DL3 file, everything before (including this IRF interpolation) needs some level of "cherenkov expert"
> 
> And this is justification for dealing with units manually instead of relying on astropy? come on.

for me it justifies (especially at the stage we are in) making it configurable from the input card if the user wants to interpolate zenith or cos zenith or HA at the price of checking the unit in contrary to having to change the code every time a different interpolation variable is tried.
Having units taken care of automatically would be preferential of course, but at least I do not know how to do it to allow the above functionality (QTable cannot do it as far as I know). Maybe general python eval can do it, but this is a pretty dangerous function. @maxnoe if you have some idea how to parse a string with fits fields names, mathematical operators and functions such to read them together with units I would very much appreciate a suggestion.

maxnoe: @jsitarek As discussed before, this also not something we want in here at all. We explicitly decided not to use some sort of configuration system. So pyirf will never have the problem of evaling a mathematical function involving column names that comes from a config file. 

I am tired of this discussion, I will come back tomorrow.

The actual unit tests  you wrote look good, the interpolation functions need some thinking what the actual input should be.

But please just remove this large example script.

jsitarek: > @jsitarek As discussed before, this also not something we want in here at all. We explicitly decided not to use some sort of configuration system. So pyirf will never have the problem of evaling a mathematical function involving column names that comes from a config file.
>
moving this out of pyirf does not solve the problem, just moves it into the pipeline implementation, so if you have some suggestion on that I would still appreciate it
 
> I am tired of this discussion, I will come back tomorrow.
we finally agree on something because I'm also tired :-)
and feeling that we are all just wasting time

> 
> The actual unit tests you wrote look good, the interpolation functions need some thinking what the actual input should be, but the approach also looks ok.
> 
> But please just remove this large example script.

I removed it and cancelled the pandas dependancy

maxnoe (changes requested):
I am not going into a deeper review here, because there are too many issues with the use of tables / units.

Since the actual interpolation code also does not need this, let me make a suggestion:

1. Let's reduce this PR to only the interpolation functions and corresponding tests (what I wrote from the beginning).
2. I will add io functions for reading the IRFs, so you can basically make round trips of `create_xxx_hdu` / `read_xxx`.

Like this you don't have to care too much about getting this astropy tables / units thing right and we will add the functionality in several PRs.

Does that sound ok to you?

So basically the only new things in this PR should be the `interpolate_effective_area` function and the `interpolate_dispersion_matrix` function. And the two tests you wrote for those.

Then these functions should be named consistently with the corresponding calculate functions, so it matches.

This means we should have `interpolate_effective_area_per_energy_and_fov` and `interpolate_energy_dispersion`.

Then the argument names should not contain any unneeded abbreviations and have a clearer meaning.

So e.g. instead of `matrix_all` use `energy_dispersions`, instead of `pars_all` I would suggest something like `grid_points` and instead of `interp_pars` maybe `target_point`?

jsitarek:
I can rename the functions and variables.

but about the IO functions, why do you want to write them yourself if I already did it? If you have some issues with them please let me know and I will correct them.

HealthyPear:
I also vouch for such code reduction and naming conventions.

Only a small comment/question: I think here `pars_all` is a bit more than only `grid_points` because is as a shape of `(n_grid_points, n_interp_dim)` so probably `grid_points_values` would be a better naming?

maxnoe:
@HealthyPear a point can be nD. Here we have a a list of nD points, so we get a 2D array of shape (n_points, n_dims).

jsitarek:
Hi @maxnoe, @HealthyPear 

I renamed the input variables and function names  using your suggestion. 
About the IO functions, I think you missed my question: why do you want to rewrite them.

From the previous discussions in this thread I think we all agree that the IRF reading functions should be a part of pyirf. Also as mentioned earlier they are rather urgently needed for application to LST. And the IO reading functions are clearly strongly related to the implemented interpolation functions. And the code that I implemented is working fine (but if you think it will fail in some cases please let me know). 

If you have any specific comments about those IRF functions (I guess you probably would like that I move them to a separate files somewhere in io directory) please let me know and I will implement those.

pyirf/interpolation.py
```diff
@@ -0,0 +1,190 @@
...
+    -------
+    aeff_interp: astropy.units.Quantity[area]
+        Interpolated Effective area array with shape (n_energy_bins, n_fov_offset_bins)
+    """
+
+    _, n_fov_offset_bins, n_energy_bins = effective_area.shape
+
+    # get rid of units
+    effective_area = effective_area.to_value(u.m**2)
+    min_effective_area = min_effective_area.to('m2').value
```
maxnoe, line 38: `to_value(u.m**2)` here as well

pyirf/interpolation.py
```diff
@@ -0,0 +1,190 @@
...
+
+    _, n_fov_offset_bins, n_migration_bins, n_energy_bins = energy_dispersions.shape
+
+    # interpolation
+    matrix_interp = griddata(grid_points, energy_dispersions, target_point, method=method)
+    matrix_interp = np.swapaxes(matrix_interp, 0, 2)
+
+    # now we need to renormalize along the migration axis
+    norm = np.sum(matrix_interp, axis=1, keepdims=True)
+    mig_norm = np.divide(matrix_interp, norm, out=np.zeros_like(matrix_interp), where=norm != 0)
```
maxnoe, line 82: I think a comment here about the usage of `out` and `where` would be good. It is a good solution, but not really obvious from the code.

So something along the lines of:

```
# By using out and where, it is ensured that columns with norm = 0 will have 0 values without raising an invalid value warning
```

pyirf/interpolation.py
```diff
@@ -0,0 +1,190 @@
...
+    return mig_norm
+
+
+def read_fits_bins_lo_hi(file_name, ext_name, tag):
```
maxnoe, line 86: If you want to implement these reading functions, Please move them to `io/gadf.py` and have one reading method per IRF component that returns all the relevant information from the HDU. 

Not one function to read the bins, one function to read the irf values and so on.

Also, again you are mixing two things, the actual IO of reading in the data and trasforming it (combining lo + hi to contiguous bins).

This is not ok.

The function to combine lo / hi to edges assuming contiguous bins (which we right now use in pyirf but are not mandated by the standard!) can go into `pyirf/binning.py`.

jsitarek: > If you want to implement these reading functions, Please move them to `io/gadf.py` 

ok, will do

> and have one reading method per IRF component that returns all the relevant information from the HDU.
> 
> Not one function to read the bins, one function to read the irf values and so on.
> 

note that the current scheme uses a single function for reading any of the different IRFs  (tested on effective area and migration matrix, but should work also fine for PSF) and the second function to read the binning, so it is not 2 functions per IRF type, but 2 functions in total (in contrary to what you propose with 1 function per IRF type), but no problem I can make wrap ups for individual IRFs that would extract the values and binning (and specify the default extension and field names for the variables passed to those functions)

> Also, again you are mixing two things, the actual IO of reading in the data and trasforming it (combining lo + hi to contiguous bins).
> 
> This is not ok.
> 
> The function to combine lo / hi to edges assuming contiguous bins (which we right now use in pyirf but are not mandated by the standard!) can go into `pyirf/binning.py`.

the transformation is from the two lists of bin edges (lo and hi) to a single list. It can be extracted to a separate function, but the same kind of transformation (but the other way around) is also used in all the functions of gadf.py: create_aeff2d_hdu, create_psf_table_hdu, create_energy_dispersion_hdu, create_background_2d_hdu, create_rad_max_hdu. 
So we should then also extract it from those functions (which might be a good idea actually, as it will avoid quite some code duplication)

maxnoe: > (which might be a good idea actually, as it will avoid quite some code duplication)

Yes!

pyirf/interpolation.py
jsitarek: done,

do you know if there is an actual change in performance between to(...).value and to_value(...), or is it just a question of style?

pyirf/interpolation.py
jsitarek: done

pyirf/io/gadf.py
```diff
@@ -63,10 +70,8 @@ def create_aeff2d_hdu(
         INSTRUME.
     """
     aeff = QTable()
-    aeff["ENERG_LO"] = u.Quantity(true_energy_bins[:-1], ndmin=2).to(u.TeV)
-    aeff["ENERG_HI"] = u.Quantity(true_energy_bins[1:], ndmin=2).to(u.TeV)
-    aeff["THETA_LO"] = u.Quantity(fov_offset_bins[:-1], ndmin=2).to(u.deg)
-    aeff["THETA_HI"] = u.Quantity(fov_offset_bins[1:], ndmin=2).to(u.deg)
+    aeff["ENERG_LO"], aeff["ENERG_HI"] = binning.split_bin_lo_hi(true_energy_bins[np.newaxis, :].to(u.TeV))
```
jsitarek, line 73: Hi @maxnoe 
I did those changes introducing dedicated functions in binning.py to split or join the bins to convert between the pyirf-like and fits-like format of binning. I will do the same change also in the other create_... functions

pyirf/interpolation.py
jsitarek: the functions reading the binning from FITS file and joining/splitting it are now separated

jsitarek:
closing this one because of PR #141

Issue #127, opened 2021-03-17
gammapy and ogadf_schema not in environment


jsitarek:
the two packages are needed by test_gadf, however they are not included in environment.yml file

HealthyPear:
hi Julian

we do not use conda right now to install pyirf, only pip

those dependencies are obtained from the dev installation as

pip install '.[all]'

or 

pip install '.[tests]' if you already installed the base package

maxnoe:
The environment is thought for users. It is not a place to specify test depedencies.

As written in the dev documentation, you can use `pip install -e .[all]` to get the dev, test and doc dependencies.
https://cta-observatory.github.io/pyirf/install.html#installing-for-development

HealthyPear:
take a look at the installation instructions if you didn't already
https://cta-observatory.github.io/pyirf/install.html

jsitarek:
thx for quick feedback. I made my install creating conda environment, so for my own copy I just added those 2 packaged to the environment file
still, I would like to understand, if the environment is for non-developers only why does it have  pytest and pytest-cov inside ?
there are also some other packages there (e.g. uproot) that I can see are used only in tests and some others (like awkward) that I cannot find a place where are used (maybe as dependancies only)

jsitarek:
I can do it

HealthyPear:
Fixed in #129

PR #128, opened 2021-03-17
added two missing packages that are needed for tests

jsitarek:
PR for  Issue #127

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -25,3 +25,5 @@ dependencies:
     - sphinx_automodapi
     - uproot4
     - awkward1
+    - gammapy
+    - ogadf_schema


maxnoe (approved):

maxnoe:
@jsitarek same story here as in lstchain, for the private test data to be available, PRs have to come from the main repo, not a fork. I pushed your branch to the main repo here: #129

jsitarek:
actually how did you set it such that the tests are running in the main repository despite the need to download the data?
maybe I could also set it up in my fork so the tests would work when I make PR from it. For this one it does not matter because it is very simple, but the other PR already has quite a few files modified, so moving the changes to a separate branch of the main project from my fork might be more tricky

maxnoe:
@jsitarek you just need to push the branch to the correct remote, nothing else is required:

```
git remote add upstream git@github.com:cta-observatory/pyirf
git push upstream interpolate_irfs
```

I tried with setting the secrets also in the fork, but it just doesn't work it seems.

HealthyPear:
> I tried with setting the secrets also in the fork, but it just doesn't work it seems.

It's because it is really not allowed on GitHub
see here https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

jsitarek:
thx @maxnoe @HealthyPear for explanations,

PR #135, opened 2021-03-18
Gammapy conversion for IRFs

maxnoe:


pyirf/io/gadf.py
```diff
@@ -85,8 +85,8 @@ def create_aeff2d_hdu(
 @u.quantity_input(
     psf=u.sr ** -1,
     true_energy_bins=u.TeV,
-    fov_offset_bins=u.deg,
     source_offset_bins=u.deg,
+    fov_offset_bins=u.deg,
```
HealthyPear, line 89: I think it was right before (is the order non-commutable here?)

PSF requires axes (energy, offset, rad) where `offset` is FoV offset and `rad` is source offset (like you wrote in the docstring)

pyirf/tests/test_gammapy_interop.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_effective_area_table_2d():
+    from pyirf.gammapy import create_effective_area_table_2d
+
+    shape = (len(true_energy_bins) - 1, len(fov_offset_bins) - 1)
+    aeff = np.random.uniform(0, 1e5, size=shape) * u.m**2
+
+    aeff_gammapy = create_effective_area_table_2d(aeff, true_energy_bins, fov_offset_bins)
+    # test str repr works
```
HealthyPear, line 21: not sure I get what this is supposed to do...

wouldn't be more robust to write a fits file HDU and read it back like the following

`aeff = EffectiveAreaTable2D.read(filename, hdu="AEFF")`

(the scope of this comment is for all the 3 test functions)

docs/gammapy.rst
```diff
@@ -0,0 +1,11 @@
+.. _gammapy:
+
+Gammapy Interoperability
+========================
+
+
```
HealthyPear, line 6: not sure if it's really that important but I would add at least one small paragraph explaining what we support in terms of gammapy classes

for example we support only PSF3D for now out of the 4 available from their API

HealthyPear (changes requested):

pyirf/io/gadf.py
maxnoe: It is just moved to have the order be the same as the argument order. No change in behaviour.

pyirf/tests/test_gammapy_interop.py
maxnoe: Why would you write a file to go from one memory-representation to another? What would be more robust about introducing a completely unnececessary round-trip through the fits format instead of just using the documented gammapy API?

docs/gammapy.rst
maxnoe: Have you actually looked at the generated docs? The three functions are immediately below this line and mention what classes are supported. It would just be repeating manually the autogenerated information.

HealthyPear: I cannot see the generated docs because they are crashing.

maxnoe: Builds fine for me.

setup.py
```diff
@@ -23,6 +23,9 @@
         "uproot~=4.0",
         "awkward~=1.0",
     ],
+    "gammapy": [
```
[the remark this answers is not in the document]
maxnoe: yes, done, was missing from the docs part.

pyirf/tests/test_gammapy_interop.py
maxnoe: They test that the conversion works. It does not assert anything, since the error case is an exception raised by gammapy (E.g. a ValueError if the axis have wrong names  or the data has wrong shape).

HealthyPear (approved):

PR #137, opened 2021-03-18
Add fact example notebook

maxnoe:
Depends on #135, so I make it a draft for now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
         with:
           docs-folder: "docs/"
           pre-build-command: |
-            apt update --yes && apt install --yes git build-essential pandoc
+            apt update --yes && apt install --yes git build-essential pandoc curl
             pip install -U pip setuptools wheel
             pip install --use-feature=2020-resolver -e '.[docs]'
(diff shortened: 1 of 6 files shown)


maxnoe:
@HealthyPear After merging #135, this is ready for review

HealthyPear (changes requested):
- the CI docs is failing to get the test data, but a local copy works
- the PSF plot is missing a legend
- the energy dispersion plot is missing the axes' labels

HealthyPear:
there aren't OFF observations to use for the bkg rate from this test data I guess

maxnoe:
@HealthyPear there would be a proton file, but the use case is not really there since for wobble observations, you estimate the background from the off regions using something like:
https://docs.gammapy.org/dev/makers/reflected.html

and not the background IRF

HealthyPear:
sure, I meant also real OFF observations, but as you say the use case is more extended than what we have here

maxnoe:
@HealthyPear real off observatins would not be part of the IRF, they are in the DL3 event list from which you select regions for background estimation.

maxnoe:
@HealthyPear Done

HealthyPear (approved):

PR #138, opened 2021-03-22
Fix name of deploy build

maxnoe:


.github/workflows/deploy.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: Deploy to PyPi
```
HealthyPear, line 1: perhaps just "Deploy" or "CD" if we want to provide also a conda package in the future?

HealthyPear (changes requested):

HealthyPear (approved):

PR #139, opened 2021-03-22
Fix how theta cut is calculated in EventDisplay comparison

maxnoe:
* Only calculate theta cut before cut optimation, do not recalculate
after
* Calculate in 5 bins per decade, interpolate to 50 bins

### DIFF
--- a/examples/calculate_eventdisplay_irfs.py
+++ b/examples/calculate_eventdisplay_irfs.py
@@ -146,7 +146,7 @@ def main():
         percentile=68,
     )
 
-    # interpolate to 50 bins be decade
+    # interpolate to 50 bins per decade
     theta_center = bin_center(theta_bins)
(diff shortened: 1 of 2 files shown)


examples/calculate_eventdisplay_irfs.py
```diff
@@ -122,36 +122,48 @@ def main():
...
         percentile=68,
     )
 
-    # same bins as event display uses
-    sensitivity_bins = add_overflow_bins(
-        create_bins_per_decade(
-            10 ** -1.9 * u.TeV, 10 ** 2.31 * u.TeV, bins_per_decade=5
-        )
-    )
+    # interpolate to 50 bins be decade
```
HealthyPear, line 149: be ---> per

HealthyPear (approved):

HealthyPear (approved):

PR #142, opened 2021-03-23
add validation function for psf norm

AMWMitchell:


HealthyPear (changes requested):
Thanks for this!

I think the code is OK, but before approving I would like this PR to have 2 small additional things,

- a unit-test under `pyirf/tests/test_validation.py` in which a PSF table not completely normalized is fed to the `validate_psf` which asserts the failure
- a section in the docs at the same level of e.g. `docs/utils.rst` called `docs/irf_validation.rst` or `docs/output_validation.rst` in which this contribution is explained - then we can use that docs page to describe any other DL3 validation tool we might introduce (you can copy/paste/edit any docs section at the same level to get the same layout)

PR #143, opened 2021-03-23
Fix zenodo json

HealthyPear:
removed some trailing commas that seem to block the automatic generation of the DOI from a released version


maxnoe (approved):
Yes, json does not allow trailing commas, sorry, should have checked in the other PR

PR #144, opened 2021-03-23
Add function to compute bias and resolution from energy dispersion

maxnoe:


pyirf/benchmarks/energy_bias_resolution.py
```diff
@@ -106,3 +106,39 @@ def energy_bias_resolution(
...
+    """
+
+    cdf = np.cumsum(energy_dispersion, axis=1)
+
+    n_energy_bins, _, n_fov_bins = energy_dispersion.shape
+
+    bias = np.zeros((n_energy_bins, n_fov_bins))
+    resolution = np.zeros((n_energy_bins, n_fov_bins))
+
+    for energy_bin in range(n_energy_bins):
```
maxnoe, line 133: Unfortunately, I didn't find a nice way of vectorizing this

HealthyPear: not really a vectorization, but you could do at least the same as you do in the unit-test but using `itertools.product`

HealthyPear (commented):
for the on-axis case we shouldn't expect big differences

I think doing this from the IRF instead of from the events themselves is more general and the energy benchmarks (meaning res and bias) would get the validation from the IRF indirectly

perhaps (apart from the ED comparison) we could think of dump the old function in favor of this one

maxnoe:
@HealthyPear I don't really see a reason to dump the old function, both are fine and needed I'd say.

pyirf/benchmarks/energy_bias_resolution.py
maxnoe: That does not really change anything but remove one level of indentation. I have no strong preference for either and it doesn't change anything for run time.

HealthyPear (approved):

pyirf/benchmarks/energy_bias_resolution.py
adonath, line 133: The only way I have found in Gammapy to avoid the loop is the "brute-force" approach of finding the closest corresponding value: compute the cdf and setup an ND interpolator, evaluate the cdf on a finer grid, with the precision you require and then use `np.argmin(cdf_fine - value, axis=)` to find the closest value. The resulting ND implementation uses a lot numpy broadcasting, see https://github.com/gammapy/gammapy/blob/master/gammapy/irf/psf/core.py#L53, for a similar problem. I'm not sure whether it is more efficient in the end, but the solution is at least ND and does not rely on the hard-coded loops over the other dimensions. Not sure I would recommend it though...

adonath: I just remember @registerrier implemented an ND solution, based on `np.nditer()` for a similar problem to avoid the hard-coded Python loops here: https://github.com/gammapy/gammapy/blob/master/gammapy/stats/counts_statistic.py#L42 Might be useful as well..

adonath: The advantage of the first solution is that it fully supports numpy broadcasting. So something like this:
```python3
containment = np.array([0.68, 80, 90])
energy_true = [1, 3, 10, 30, 100] * u.TeV
offset = [0.5, 1] * u.deg

psf.containment_radius(
    containment=containment.reshape((-1, 1, 1)),
    energy_true=energy_true.reshape((1, -1, 1)),
    offset=offset.reshape((1, 1,- 1))
)
```
Returns the expected broad-casted result. Not sure it's need here but in general it's nice.

maxnoe: @adonath Thanks a lot!

PR #145, opened 2021-03-29
Proton+Helium spectrum from DAMPE 2019 ICRC proceeding

moralejo:
Added a proton+Helium spectrum from DAMPE (unfortunately I could only find a proceeding as reference for it: https://inspirehep.net/files/62efc8374ffced58ea7e3a333bfa1217).
It is useful for some real background data vs. MC proton comparisons in terms of absolute rates, since the data before g/h separation cuts contain mostly p and He. One should anyway keep in mind that He and proton showers (and hence images) differ, which is why  residual hadronic  background after g/h separation is mostly made of protons.


maxnoe:
@moralejo thanks, that looks useful. However, I would probably go for a either a function or a interpolation class / instance instead of this lambda function. 

If you are fine with it, I will push the changes to this branch. Otherwise I can also open a PR targetting this branch.

maxnoe:
@moralejo Is this data available here maybe?

https://lpsc.in2p3.fr/crdb/

moralejo:
> @moralejo Is this data available here maybe?
> 
> https://lpsc.in2p3.fr/crdb/

No, I had tried that database already, but unsuccessfully: no He data for DAMPE or CALET. There is no refereed publication on this, perhaps that is the problem. There are He data from space detectors (like AMS), but at lower energies, or from balloons at the energies of our interest (ATIC, CREAM, Nucleon-KLEM). The p+He from the balloon experiments are in the same ballpark (within 15% or so) of the DAMPE spectrum from the ICRC proceeding.

moralejo:
The overlaid thin like is the interpolation implemented here:

comparison

moralejo:
Thanks @maxnoe for the improvements. I just removed quite a few unnecessary digits that I had included in the flux values, just to make it look a bit nicer. 
Fine for me to move on.

HealthyPear (changes requested):
Since there is no proper test available for these changes, I would also update `examples/plot_spectra.py` (superimposing also the data points from DAMPE) and related documentation at the bottom of `docs/examples.rst`

maxnoe:
@HealthyPear I am adding a test right now

HealthyPear (changes requested):
Only small thing missing is to update `docs/examples.rst`.

I also suggested to add the DAMPE points to show the level of interpolation, but probably it's a bit overkill (in that case we should do that for all curves and it seems too much).

maxnoe:
>  (in that case we should do that for all curves and it seems too much).

The other curves are parametrized functions, there are no points to show.

HealthyPear:
> The other curves are parametrized functions, there are no points to show.

Sure, but they are still representing real data which can be found in the CR database if we want.
Anyway, I have no strong feelings about this, so for me, it's OK to show only the functions.

maxnoe:
> Sure, but they are still representing real data which can be found in the CR database if we want.

No. We directly use the published (somewhere, not necessary a proper publication) parametrizations for all the others. The HEGRA / MAGIC spectra are the result of forward-folding, so there are no flux points corresponding to the spectrum.  

Interpolation is obviously different to fitting and we directly use the points.

HealthyPear (approved):

HealthyPear:
Today this popped out
https://arxiv.org/abs/2105.09073

PR #148, opened 2021-04-09
Use setuptools_scm for versioning

maxnoe:


HealthyPear (changes requested):
I think you should update the CI workflow like it has been done in ctapipe, i.e.

```
- uses: actions/checkout@v2
        with:
          fetch-depth: 0
```

maxnoe:
@HealthyPear Good catch, done

HealthyPear (approved):

PR #149, opened 2021-04-26
Interpolation docs

maxnoe:


docs/interpolation.rst
```diff
@@ -0,0 +1,17 @@
+.. _interpolation:
+
+Interpolation of IRFs
+=====================
+
+This module contains functions to interpolate from a set of IRFs for different
+condiditions to a new IRF.
```
HealthyPear, line 7: small typo

docs/interpolation.rst
```diff
@@ -0,0 +1,17 @@
...
+
+This can e.\,g. be used to interpolate IRFs for zenith angles of 20° and 40°
+to 30°.
```
jsitarek, line 10: Hi @maxnoe 

I think it should me mentioned in the documentation that the interpolation is done over two dimensions (e.g. zenith and azimuth).

And sorry for forgetting the old compare_irf_cuts function - thx for fixing it here.

maxnoe: It *could* be done using two dimensions, however, the functions allow arbitrary number of dimensions, right?

docs/interpolation.rst
```diff
@@ -0,0 +1,17 @@
+.. _interpolation:
+
+Interpolation of IRFs
+=====================
+
+This module contains functions to interpolate from a set of IRFs for different
+condiditions to a new IRF.
+
+This can e.\,g. be used to interpolate IRFs for zenith angles of 20° and 40°
```
HealthyPear, line 9: why the comma?

HealthyPear (changes requested):

docs/interpolation.rst
maxnoe: Sorry, LaTeX muscle memory. This inserts a small non-breaking space in LaTeX, which is the proper way to typeset abbreviations.

docs/interpolation.rst
HealthyPear: is "IRFsaa" a special thing I don't know or a typo?

maxnoe: typo :see_no_evil:

jsitarek: Hi @maxnoe 
bulk of the functions is done in a way that it would allow an arbitrary number of dimensions, but the actual function used for the interpolation: griddata I think works only in 2D. Since this is the most typical use case, I think this is not a problem, but we can also consider substituting the griddata function with some other interpolation engine for higher number of dimensions.

pyirf/spectral.py
```diff
@@ -242,6 +245,8 @@ class TableInterpolationSpectrum:
     By default, flux is interpolated linearly in log-log space.
     '''
     def __init__(self, energy, flux, log_energy=True, log_flux=True, reference_energy=1 * u.TeV):
+        '''Create a new TableInterpolationSpectrum spectrum'''
+        super().__init__(normalization=normalization, index=index, e_ref=e_ref)
```
jsitarek, line 249: this does not seem to be related to the PR, and also the commit says only "Add docstring for __init__ methods ", while this last line does something different, was it intended ?

maxnoe: Good catch, I unintentionally copied two lines.

docs/interpolation.rst
maxnoe: https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html

No, griddata does arbitrary number of dimensions

maxnoe:
Mmh. I cannot reproduce the three warnings locally

maxnoe:
Ha, now it works.

HealthyPear (approved):

HealthyPear (approved):

Issue #126, opened 2021-03-01
IRF output files should have sufficient metadata


kosack:
When writing IRF files in `pyirf.io.gadf` format, it should be possible to add some extra metadata describing the context of the IRF and its  observation parameters.

That will make it easier later to open and interpolate between IRFs.   For that we should define an initial set of header keywords.  Perhaps, reusing some standard FITS headers when possible, such as:

```
IRFTYPE  = "fixed" / IRF type ("fixed" bin in phase space or "averaged" over obervation conditions)
ALTITUDE = 80.0    / altitude of simulation in degrees (zenith is then 90-ALTITUDE) [deg]
AZIMUTH  = 180.0   / azimuth of simulation [deg] 
NSB      = 1       / NSB level  multiplier (1=1x nominal)
```

Additionally, having the layout name, subarray name, prod name, etc would help avoid errors.  

This will facilitate the bookkeeping that is necessary to do IRF interpolation, as in #124 #125

maxnoe:
This is already possible,  you can pass arbitrary header cards to the create HDU functions or edit the headers afterwards.

kosack:
This should be in the example notebooks then, as a best practice.  Otherwise, it will be done in a non-standard way by people that use the library

maxnoe:
I am very reluctant on adding any more stuff to the event display example. Even more, I'm inclined to completely remove it as an example and only have it in the tests.

It is not an example on how pyirf should be used, it was just a script to compare the results to Event Display.

People using pyirf should understand what the functions do and what the needed input / the output is. Not just follow this "example".

I also don't want to duplicate all the optional headers of the GADF specification or even add new conventions. 

Or we need to rediscuss the scope of pyirf. In my understanding, pyirf should follow a specification (right now GADF) not create a new specification.

jsitarek:
Hi @kosack the headers will be tricky, because it is difficult to know what we will want to have inside them in one year from now or so. Azimuth and Altitude sounds fine, but again it might happen that some simulations are done for points in the sky, while others for small ranges of Alt/Az. Also we should be very clear which azimuth we take, the Corsika one, or the geographical one, and if the range is -180 to 180 or 0-360
[...]

PR #150, opened 2021-04-26
added the IO functions removed from PR #141

jsitarek:
Since #141 is already merged, as announced there I moved into this separate PR the IO functions. 
the summary of the discussion with @maxnoe from the previous PR was that Max was not in favor of having functions that read arrays of IRFs, while I think this is very useful because it is not trivial as some transformations have to be done (separate lo/high bins vs a single list of bins, change of sequence of axes) and they are common in all the IRFs, so if not implemented this way, would result in quite some code repetition. There are some two generic functions in this PR: read_irf_grid and read_fits_bins_lo_hi that read in the actual IRFs (whatever type, including the transformation), and bins values (including their consistency), and two wrapper functions for Aeff and migration matrix. 

Please let me know of any modifications of those functions that you would like

### DIFF
--- a/pyirf/io/gadf.py
+++ b/pyirf/io/gadf.py
@@ -4,6 +4,7 @@ from astropy.io.fits import Header, BinTableHDU
 import numpy as np
 from astropy.time import Time
 import pyirf.binning as binning
+from pyirf.cuts import compare_irf_cuts
 
 from ..version import __version__
 
@@ -13,6 +14,11 @@ __all__ = [
     "create_energy_dispersion_hdu",
     "create_psf_table_hdu",
     "create_rad_max_hdu",
+    "compare_irf_cuts_in_files",
+    "read_fits_bins_lo_hi",
+    "read_irf_grid",
+    "read_aeff2d_hdu",
+    "read_energy_dispersion_hdu",
 ]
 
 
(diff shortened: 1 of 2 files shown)


pyirf/io/gadf.py
```diff
@@ -303,3 +309,201 @@ def create_rad_max_hdu(
...
+    migration_bins: numpy.ndarray
+        Bin edges for the relative energy migration (``reco_energy / true_energy``)
+    fov_offset_bins: astropy.units.Quantity[angle]
+        Bin edges in the field of view offset.
+    """
+
+    field_name = "MATRIX"
+    energy_dispersion = read_irf_grid(file_name, extname, field_name)
+    last_axis = len(energy_dispersion.shape) - 1
+    energy_dispersion = np.swapaxes(energy_dispersion, last_axis - 2, last_axis)
```
maxnoe, line 503: Why swapping axis? That looks wrong too me. The only thing needed should be the transpose when reading from the table, not any swapaxes stuff.

jsitarek: I think it is the same - transposition is exactly swapping axes. Please note that this is a 4D array, and you only need to swap two of them. I guess  .transpose(0, 1, 3, 2) would also work here, but swapaxis looks more clear to me here.

maxnoe: No. The point is, transposing without any arguments should be done! Not just swapping two axis.

Could it be that the file your are testing this on does not follow the "Recommended axis order"?

jsitarek: the test is done on prod 3 MCs, but as far as I remember it had the same axis sequence as the files saved  with pyirf.
Nevertheless, as we discussed in e-mail, I will update the tests to make them round tests so I will check it again.
Nevertheless I'm sure that it cannot be simple transpose, it would turn the axes 0, 1, 2, 3 into 3, 2, 1, 0 and this would not be correct.

maxnoe: The recommended axis order is `ENERGY, MIGRA, THETA`, which is what we do here:

https://github.com/cta-observatory/pyirf/blob/0e7fcd29297a9ae95910f736bddcb73bb9cc1066/pyirf/irf/energy_dispersion.py#L67-L70

And we simply transpose to meet FITS conventions when exporting the HDU here:

https://github.com/cta-observatory/pyirf/blob/0e7fcd29297a9ae95910f736bddcb73bb9cc1066/pyirf/io/gadf.py#L185

pyirf/io/gadf.py
```diff
@@ -303,3 +309,201 @@ def create_rad_max_hdu(
...
+    if isinstance(files, str):
+        files = [files]
+
+    n_files = len(files)
+
+    irfs_all = list()
+
+    for this_file in files:
+        # [0] because there the IRFs are written as a single row of the table
+        irfs_all.append(QTable.read(this_file, hdu=extname)[field_name][0])
```
maxnoe, line 437: The transpose is missing here, this is why you have to do non-trivial transposes in edisp.

This goes back to the point that I really don't understand why we need this `read_irf_grid` function. Simple functions that read a table once and return the arrays. And stacking can then just be `edisp = np.stack(edisps)`

maxnoe: Also, keeping the table around is good, since it contains the metadata. So this function is just not as general as you might think it is.

pyirf/io/gadf.py
jsitarek: Thx @maxnoe indeed there was a problem, but not with migration matrix (which was read correctly), but with effective area, where messed up the axis order during refactoring of the code. Now I do the .T in read_irf_grid and it works fine for both Aeff and migration matrix

pyirf/io/gadf.py
jsitarek: I switched to np.stack. 
Please note @maxnoe that the function is also making the required transposition that is there for all the IRFs and extracting the 0-th row - all the operations to convert between the FITS format and the one used inside pyirf. And it is working fine on either individual files or lists of files so I think it is very useful and it makes the custom functions that read individual IRFs simpler. 

I agree that the table includes additional important information via metadata, but I think it would be more useful to extract this metadata and return such concrete information. Nevertheless, first we would need to agree (in #126 ) what information should be put there, with what names, etc.

maxnoe: @jsitarek 

> Nevertheless, first we would need to agree (in #126 ) what information should be put there, with what names, etc.

Actually no. Since at no point pyirf will use this metadata, users of pyirf can fill and read any metadata they like and use it for whatever purpose. 

That's also the problem with this `read_irf_grid` function. You would have to open all the files again to also read the metadata.

So, please, write functions that read each IRF type, validating it's the correct type by looking at HDUCLASX header keywords and returning also the metadata.

This functions can than easily be called for multiple files and the results stacked / compared.

I think this has much better advantages than the little code repetition you avoid by  using this read_irf_grid function.

pyirf/io/gadf.py
```diff
@@ -303,3 +308,206 @@ def create_rad_max_hdu(
...
+    if isinstance(files, str):
+        files = [files]
+
+    cuts = list()
+    for file_name in files:
+        table = QTable.read(file_name, hdu=extname)
+        cuts.append(table)
+
+    # if the function is run on single file do not need the first axis dimension
+    if len(files) == 1:
```
maxnoe, line 356: If I give a list with one file as input, I want a list with one entry back.

jsitarek: done

pyirf/io/gadf.py
jsitarek: > @jsitarek
> 
> > Nevertheless, first we would need to agree (in #126 ) what information should be put there, with what names, etc.
> 
> Actually no. Since at no point pyirf will use this metadata, users of pyirf can fill and read any metadata they like and use it for whatever purpose.
> 
well, I would not be so sure about this. If we really have standardized metadata we can use them inside pyirf to do e.g. standarized interpolation over those values. 

> That's also the problem with this `read_irf_grid` function. You would have to open all the files again to also read the metadata.
> 
as I explained above the metadata processing can be also added to this function. Please note that they should be likely the same for all the types of IRFs (no matter if you want to read Aeff or energy migration you still want to know at which zenith it was produced) so it takes perfect sense to read them in standarized way

> So, please, write functions that read each IRF type, validating it's the correct type by looking at HDUCLASX header keywords and returning also the metadata.
> 
I just implemented checking of all the HDUCLASX headers. 
Again because of using one universal function it was pretty easy, especially that many of those fields are the same for different IRFs.


> This functions can then easily be called for multiple files and the results stacked / compared.
> 
> I think this has much better advantages than the little code repetition you avoid by using this read_irf_grid function.

With every thing that we discuss in this thread the amount of code repetition in the approach that you suggested would grow larger and larger. Already now we have 3 issues that are repeated for all IRFs: 
- swapping of axis sequence 
- checks of  HDUCLASX headers
- combining IRFs produced for different parameters

and we have two IRFs already implemented, and a few more that can be added easily. So in the end I think we are avoiding a lot of code repetition

jsitarek:
Hi @maxnoe 

Coming back to this old story of the I/O classes. I implemented your comments back in May. We have a difference of opinion about the code repetition/generic methods, please check my latest replies from May, and let's see if we can converge and merge the I/O classes.

maxnoe:
@jsitarek I think we should revisit this in a more global refactoring, now that gammapy was chosen as science tool for cta. 

I think it makes sense to use the gammapy data structures inside pyirf, that will make many things a lot easier.

As for the IO parts, LST already is using gammapy, so you should also just use gammapy to load the irfs.

jsitarek:
Hi @maxnoe, ok, so what is the proposal? You want to close the IO PR? or leave it open, but without updates for the time being?
After the refactoring merging would be messy, and if we agree on this that pyirf should have some IO classes inside I think it would be good to have those inside during refactoring, because there are all the tests of individual writing functions already implemented here, and hence when you refactor you would get immediately consistency tests done. 

As we were discussing some months ago, all those IRF developements were urgent in LST, hence the IO was implemented separatelly, but if the IO functions are inside pyirf, lstchain would definitely profit from those

PR #156, opened 2021-05-24
Interpolate psf

jsitarek:
a function to interpolate PSF tables. I made it in a similar style like the other two interpolation functions. One difference is that I added an option to  interpolate over cumulative distribution. In principle this could help in the binning is fine and thus event statistics are sparse. For the tests that I did I haven't seen any significant difference, but the performance might depend on a particular use case. The default is to interpolate over differential distribution (which is slightly simpler)

@maxnoe I incorporated your comments about solid cone function, please let me know if you have any further comments

### DIFF
--- a/pyirf/tests/test_interpolation.py
+++ b/pyirf/tests/test_interpolation.py
@@ -31,6 +31,22 @@ def test_interpolate_effective_area_per_energy_and_fov():
     assert np.allclose(aeff_interp[:, 0], aeff0, rtol=0.03, atol=min_aeff)
 
 
+def calc_mean_std(matrix, vals):
+    """Auxiliary function to compute mean and std from 'matrix' along an axis which values are in 'values'."""
+    n_en = matrix.shape[0]
+    means = np.empty(n_en)
+    stds = np.empty(n_en)
+    for i_en in np.arange(n_en):
+        w = matrix[i_en, :]
+        if np.sum(w) > 0:
+            means[i_en] = np.average(vals, weights=w)
+            stds[i_en] = np.sqrt(np.cov(vals, aweights=w))
+        else:  # we need to skip the empty columns
+            means[i_en] = -1
+            stds[i_en] = -1
+    return means, stds
+
+
 def test_interpolate_energy_dispersion():
     """Test of interpolation of energy dispersion matrix using a simple dummy model."""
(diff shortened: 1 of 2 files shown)


jsitarek:
Hi @maxnoe 

can you approve the PR? as I wrote above, your comments are already implemented

pyirf/tests/test_interpolation.py
```diff
@@ -31,6 +31,22 @@ def test_interpolate_effective_area_per_energy_and_fov():
...
 
 
+def calc_mean_std(matrix, vals):
+    """Auxiliary function to compute mean and std from 'matrix' along an axis which values are in 'values'."""
+    n_en = matrix.shape[0]
+    means = np.empty(n_en)
+    stds = np.empty(n_en)
+    for i_en in np.arange(n_en):
+        w = matrix[i_en, :]
+        if np.sum(w) > 0:
```
maxnoe, line 41: For weights to make sense, they have to be all > 0, not just the sum

maxnoe (approved):

maxnoe:
Looks good to me, sorry for the delay @jsitarek

pyirf/tests/test_interpolation.py
jsitarek: actually there should be no negative weights, but it can happen at low energies that all bins are empty and you have 0s everywhere, this check is meant exactly for such a case

jsitarek:
thx for quick feedback @maxnoe, I merged the PR

maxnoe:
I wanted to give @HealthyPear a chance to review as well, but I think it's ok

HealthyPear:
I was about to re-start, but I understand it was open since a while...
Anyway at first sight I have nothing against -  just wondering if it made sense to make 2 assertions within the same test, one for high counts and one for sparse ones, so you could check if the weights check was right

jsitarek:
Hi, sorry @HealthyPear, since there were no comments from you before in this PR I did not think that you wanted to still have a look into this. If you have any further comments, please let me know and I can make a separate PR with them. 
About what you wrote above, I'm not sure if I understand it right. When I check the normalization I do an assertion with and 'or', if the bins normalize to 1 or to 0, the second case is when there is a completely empty row (because of lack of surviving MC at some energy)

PR #162, opened 2021-09-02
Refactor hist normalization, remove assert from library code

maxnoe:
I saw we had a defensive assert in the actual code, which is actually covered by the unit tests.

I also found a more understandable way of doing the normalization than this swapaxes stuff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,23 +50,31 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: git fetch --tags
+        with:
+          fetch-depth: 0
 
-      - name: Download data
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install doc dependencies
+        run: |
+          sudo apt update --yes && sudo apt install --yes git build-essential pandoc curl graphviz
+          pip install -U pip setuptools wheel
+          pip install -e .[docs]
+          git describe --tags
+          python -c 'import pyirf; print(pyirf.__version__)'
+
+      - name: Download data and run example
         env:
           DATA_PASSWORD: ${{ secrets.DATA_PASSWORD }}
         run: |
           ./download_test_data.sh
+          python examples/calculate_eventdisplay_irfs.py
 
-      - uses: ammaraskar/sphinx-action@master
-        with:
-          docs-folder: "docs/"
-          pre-build-command: |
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -50,23 +50,24 @@ jobs:
...
+          python-version: 3.8
+
+      - name: Install doc dependencies
         run: |
-          ./download_test_data.sh
+          sudo apt update --yes && sudo apt install --yes git build-essential pandoc curl graphviz
+          pip install -U pip setuptools wheel
+          pip install -e .[docs]
+          git describe --tags
+          python -c 'import ctapipe; print(ctapipe.__version__)'
```
HealthyPear, line 67: `ctapipe` should be changed in `pyirf` (this is also the reason why the `docs` part of the CI is failing)

HealthyPear (changes requested):
Code is fine, there is only a fix to be done in the CI

.github/workflows/ci.yml
maxnoe: thx

.github/workflows/ci.yml
```diff
@@ -50,23 +50,24 @@ jobs:
...
-        with:
-          docs-folder: "docs/"
-          pre-build-command: |
-            apt update --yes && apt install --yes git build-essential pandoc curl
-            pip install -U pip setuptools wheel sphinx
-            pip install -e '.[docs]'
-            python examples/calculate_eventdisplay_irfs.py
-          build-command: make html SPHINXOPTS="-W --keep-going -n --color -w /tmp/sphinx-log"
+      - name: Build docs
+        run: make doc SPHINXOPTS="-W --keep-going -n --color -j auto"
```
HealthyPear, line 70: sorry I overlooked this: we don't have a MakeFile with the definition for `doc`, so just `make html` will work

HealthyPear (changes requested):

maxnoe:
@HealthyPear Sorry I let this slip for so long, it should be fine now

setup.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 
-gammapy = "gammapy~=0.18"
+gammapy = "gammapy~=0.18.0"
```
HealthyPear, line 5: Not a problem, but what's the need for this variable substitution?

maxnoe: We have `gammapy` in three optional dependencies. This is to make sure, we use the same version everywhere. I just forgot it at one point (used the variable in 2 and the string in another one)

HealthyPear (approved):

Issue #161, opened 2021-08-10
Necessary improvements to IRF interpolation (Handle PDFs properly, extrapolation)


maxnoe:
In the current interpolation code, the amplitude of the PDF is interpolated. I think I saw a double bump structure in the energy dispersions shown by @jsitarek and checked what happens when interpolating between two gaussians where the mean and the std dev depends linearly on the interpolation variable. (E.g. mean and std get larger with a hidden variable `z`).
This is I think the expected case for e.g. energy dispersion vs. zenith.

The results show the double bump structure and make it pretty clear that interpolating the height of the pdf is not really a good solution:

```python
import numpy as np
from scipy.stats import norm
import matplotlib.pyplot as plt
from scipy.interpolate import interp1d


def expected_mean(z):
    return 5 + z / 10

def expected_std(z):
    return 1 + z / 30

x = np.linspace(0, 20, 1000)
zs = {
    z: norm(expected_mean(z), expected_std(z)).pdf(x)
    for z in [20, 40, 60]
}
y = np.array([zs[20], zs[60]]).T

f = interp1d([20, 60], y)
    
plt.figure(constrained_layout=True)
for z, pdf in zs.items():
    plt.plot(x, pdf, label=f'Truth, z={z}')
    
plt.plot(x, f(40), label='Interpolated z=40')
plt.legend()
```

Result:

![norm_interpolation](https://user-images.githubusercontent.com/5488440/128872031-32b2ac63-3159-4e75-9bce-9098b2b93aed.png)

So we should probably reconsider the approach for psf / edisp interpolation. We also should add a warning to the docs that the interpolation methods are experimental and not yet verified to work in real analyses.

jsitarek:
Hi @maxnoe 

The accuracy and the possible features will depend strongly on what we are interpolating. The current code is a pretty simple approach of amplitude interpolation, but I wanted first something simple that would not be dependent on assumptions, especially that at the time of the development we had no data to test it properly.
Now that we have more MCs that were generated, I can look into more clever approaches. One possibility that I was thinking about is to compute the first and central second moments normalize the distributions to them, and then interpolate the values of those moments. This would work much better (actually even give exactly correct solution) in the case that you quoted, but a technical issue is that you need some base shape of the distribution (I could use the closest point in the grid for this). 

I will have a look and let you know. Nevertheless, even if a more clever solution is implemented, I think it would be also good to keep an option to run with the original code.

moralejo:
[...]

PR #163, opened 2021-10-19
Implementation of a quantile PDF interpolation Procedure for E-Disps

RuneDominik:
Implements the Quantile PDF Interpolation proposed by @moralejo from [this paper](https://engineering.ucsc.edu/sites/default/files/technical-reports/UCSC-SOE-13-13.pdf) for the interpolation of energy dispersion with one hidden variable. Adresses #161

### DIFF
--- a/pyirf/interpolation.py
+++ b/pyirf/interpolation.py
@@ -169,6 +169,7 @@ def interp_hist_quantile(
     # determine quantiles step
     percentages = np.arange(0, 1 + 0.5 * quantile_resolution, quantile_resolution)
     mids = bin_center(edges)
+
     quantiles = np.apply_along_axis(numerical_quantile, axis, hists, mids, percentages)
 
     # interpolate quantiles step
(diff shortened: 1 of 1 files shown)


RuneDominik:
The latest commit implements the usage of the method for 2D grids by chaining 1D interpolations on a grid decomposed into triangular cells by a Delaunay transformation. This chaining might however propagate interpolation errors in an unpredictable manner and should therefore be tested carefully. The interpolation error does, not surprisingly, rise with the distance between target- and grid-points, as can be seen in the quick test I've done below.

[2D_reference_interpolation.pdf](https://github.com/cta-observatory/pyirf/files/7462993/2D_reference_interpolation.pdf)

pyirf/interpolation.py
```diff
@@ -2,22 +2,239 @@
...
+    cdf = np.cumsum(pdf)
+    interp = interp1d(
+        cdf / cdf.max(), mids, fill_value=(mids.min(), mids.max()), bounds_error=False
+    )
+
+    # Lookup ppf value
+    return interp(percentile)
+
+
+def rebin(entries, mids, width):
```
adonath, line 82: An elegant solution for this problem is to use https://numpy.org/doc/stable/reference/generated/numpy.ufunc.reduceat.html, this way you get ND support and can avoid the Python for loop. However for rebinning along a single axis, the Python for loop might be performant enough. You can use it with arbitrary bins size, just defined by the bin indices.

pyirf/interpolation.py
```diff
@@ -2,22 +2,239 @@
...
+    # positions q_bar.
+    width = np.diff(edges)
+    helper = np.concatenate((q_bar, V_bar), axis=axis)
+    interpolated_histogram = np.apply_along_axis(rebin, axis, helper, mids, width)
+
+    # Re-Normalize, as the normalisation is lost due to approximate nature of this method
+    # Set norm to nan for empty histograms to avoid division through 0
+    if normalize == "sum":
+        norm = np.sum(interpolated_histogram, axis=axis)
+        norm = np.repeat(
```
adonath, line 214: I think there is now need to repeat the values here, you can just rely on numpy broadcasting.

RuneDominik:
The #174 version should be sufficient for now and is much leaner. If we find that this "original" method has advantages we can still revisit this.

maxnoe:
Ok, great, then I'll go ahead and make the release

PR #164, opened 2021-11-26
Modified gammapy.py, now working for me with gammapy 0.19

mdpunch:
Hi @maxnoe ,
I've got this working with my code (derived from the FACT example) and gammapy 0.19.
I tested dumbly by putting the revised gammapy.py as a cell in my jupyter notebook, and getting the functions from there instead of importing.  I guess there is some way to modify the library and reload, but I'm not familiar with the machinery.
Let me know if there are other steps I should be taking...
Good Luck,
  Michael.
p.s. I guess this won't work with gammapy < 0.19 any more, I'm not sure how you handle that.


maxnoe:
@mdpunch thanks for the contribution.

> I guess this won't work with gammapy < 0.19 any more, I'm not sure how you handle that.

Please adapt the required version in `setup.py` to `gammapy~=0.19` and in `environment.yml` to `gammapy=0.19`

> I tested dumbly by putting the revised gammapy.py as a cell in my jupyter notebook, and getting the functions from there instead of importing. I guess there is some way to modify the library and reload, but I'm not familiar with the machinery.

The easiest is something like "Restart Kernel and Run All". If you don't want to restart the kernel, you can use `importlib.reload` like this:

Somewhere up:
```
from pyirf import gammapy as pyirf_gammapy
from importlib import reload
```

And then when a reload is required:

```
pyirf_gammapy = reload(pyirf_gammapy)
```

This only works on module level, not if you import single functions like this: `from pyirf.gammapy import ...`

maxnoe:
@mdpunch But actually the best method is just to run the unit tests (first maybe only for the functions you changed, then for everything else to make sure you didn't break something by accident / everything else also works with the new gammapy version):

```
pytest -v pyirf/tests/test_gammapy.py
```

mdpunch:
> > I guess this won't work with gammapy < 0.19 any more, I'm not sure how you handle that.
> 
> Please adapt the required version in `setup.py` to `gammapy~=0.19` and in `environment.yml` to `gammapy=0.19`
Done (not uploaded yet).
> 
> > I tested dumbly by putting the revised gammapy.py as a cell in my jupyter notebook, and getting the functions from there instead of importing. I guess there is some way to modify the library and reload, but I'm not familiar with the machinery.
> 
> The easiest is something like "Restart Kernel and Run All". If you don't want to restart the kernel, you can use `importlib.reload` like this:
> 
> Somewhere up:
> 
> ```
> from pyirf import gammapy as pyirf_gammapy
> from importlib import reload
> ```
> 
> And then when a reload is required:
> 
> ```
> pyirf_gammapy = reload(pyirf_gammapy)
> ```
> 
> This only works on module level, not if you import single functions like this: `from pyirf.gammapy import ...`

Thanks, my problem though is that I wasn't importing the modified module at all.
I have cloned the repository, and if I run python/ipython interpreter, this works fine
```
>>> import pyirf
>>> pyirf.__file__
'/scr/punch/pyirf/pyirf/__init__.py'
```
But in Jupyter/Spyder, this works also if I create a new notebook, but not if I open `docs\notebook\fact_example.ipynb`.

I'd like to update that notebook too, of course, to keep things in sync, so it would be nice if I could get the modified import to work, instead of it using the one in the miniconda env.

Pointers would be appreciated.

maxnoe:
> Thanks, my problem though is that I wasn't importing the modified module at all.

I don't understand. How could you use the code without importing? I don't mean your later try where you copied the code into the notebook.

mdpunch:
> 



> > Thanks, my problem though is that I wasn't importing the modified module at all.
> 
> I don't understand. How could you use the code without importing? I don't mean your later try where you copied the code into the notebook.

Sorry, I mean, importing from the wrong place.
> 
> Please see https://cta-observatory.github.io/pyirf/contribute.html
> 
> Edit, actually more relevant is: https://cta-observatory.github.io/pyirf/install.html#installing-for-development

Will look at those links now.

In the meantime, adding 
```
import sys
sys.path.insert(0,'/scr/punch/pyirf')
print(sys.path)
import pyirf
pyirf.__file__
```
... at the top of the FACT notebook allowed me to execute and debug it, but it's obviously not ideal.

I'll check out those links now.  Thanks!

Will re-submit when done.

P.S. the pytests run fine for the `test_gammapy_interop.py`

maxnoe:
@mdpunch You need to pip install the module locally in "editing mode", Preferably,  using a python virtualenv or conda env.

If you have a conda already installed, in the base of this repository:

```
$ conda env create -f environment.yml
$ conda activate pyirf
$ pip install -e .
```

maxnoe:
@mdpunch please clear the notebook before committing, this makes it possible to see the actual changes to the code: Cell -> Clear all outputs

mdpunch:
> @mdpunch You need to pip install the module locally in "editing mode", Preferably, using a python virtualenv or conda env.
> 
> If you have a conda already installed, in the base of this repository:
> 
> ```
> $ conda env create -f environment.yml

This gives:
```
ollecting package metadata (repodata.json): done
Solving environment: done

Downloading and Extracting Packages
astropy-4.3.1        | 7.7 MB    | ################################################################################################################################################################################################ | 100% 
numpy-1.21.4         | 6.1 MB    | ################################################################################################################################################################################################ | 100% 
sphinx_rtd_theme-1.0 | 2.7 MB    | ################################################################################################################################################################################################ | 100% 
docutils-0.17.1      | 760 KB    | ################################################################################################################################################################################################ | 100% 
Preparing transaction: done
Verifying transaction: done
Executing transaction: \ Enabling notebook extension jupyter-js-widgets/extension...
      - Validating: OK

done
Installing pip dependencies: | Ran pip subprocess with arguments:
['/scr/punch/miniconda3/envs/pyirf/bin/python', '-m', 'pip', 'install', '-U', '-r', '/scr/punch/pyirf/condaenv.s_cffw0p.requirements.txt']
Pip subprocess output:

Pip subprocess error:
ERROR: Invalid requirement: 'gammapy=0.19' (from line 5 of /scr/punch/pyirf/condaenv.s_cffw0p.requirements.txt)
Hint: = is not a valid operator. Did you mean == ?

failed

CondaEnvException: Pip failed
```

mdpunch:
> @mdpunch You need to pip install the module locally in "editing mode", Preferably, using a python virtualenv or conda env.
> 
> If you have a conda already installed, in the base of this repository:
> 
> ```
> $ conda env create -f environment.yml
> ```

I don't understand.  Doing this gives me a new conda env, `pyirf`, and the paths for the modules are `/scr/punch/miniconda3/envs/pyirf/...`.  Then editing my git fork doesn't do anything, since those aren't the modules loaded...

Seems to me a better solution to set the pythonpath to override local installations: 
```export PYTHONPATH='/scr/punch/pyirf'```

Doing this allows me to edit the local fork, and then do commits, pushes, etc. (works for me, at least).

Am I missing something?  Does installing `pip install -e '.[all]'` put in links my git fork, or something?

**Comment by maxnoe** [comment]
> Am I missing something? Does installing pip install -e '.[all]' put in links to my git fork, or something?

Yes, the `-e` switch is for `editable` also called "Developer Mode". It installs symlinks from the current directory (the `.` in the command) into the python in used. So you create a new conda env with all needed dependencies and symlinks to the development repository for pyirf.

> Seems to me a better solution to set the pythonpath to override local installations:
export PYTHONPATH='/scr/punch/pyirf'

No, for many reasons, messing around with PYTHONPATH is almost always the worst possible solution. Same goes for tinkering with `sys.path` from within python.

> Pip subprocess error:
ERROR: Invalid requirement: 'gammapy=0.19' (from line 5 of /scr/punch/pyirf/condaenv.s_cffw0p.requirements.txt)
Hint: = is not a valid operator. Did you mean == ?

Sorry, I forgot that gammapy is in the `pip` section of the conda environment as there were problems with installing it from conda-forge at the time. Please use the same syntax as in the setup.py for now `~=0.19.0` meaning compatible to 0.19 and at least `0.19.0`

**Comment by mdpunch** [comment]
Okay, thanks for the explanations !

Maybe I'll do a fork to add that to the docs, since it could help future devs ?

**Comment by mdpunch** [comment]
> Or I could add this editable version to the current environment with: `conda env update --name my_env_name --file environment.yml` ... unless there is some drawback to that?
> 
> (the reason my wanting to do this is that in the `my_env` I have other things that I might use, like jupytext, jupyterlab, jupyter_nbextensions_configurator, h5py, dill... but if this should be avoided I'm fine with that, if I can understand why...)

But I guess it's cleaner to have a dedicated `pyirf` env for the editing and testing (that it would be cleaner), but that I could then use my version of the `pyirf` in other environments by installing using (from the pyirf directory):
```conda env update -f environment.yml``

**Comment by maxnoe** [comment]
> I guess I don't have to create a completely new conda env, do the pip command in the env in which I want to test things out?
(edit: maybe not, since the environment.yml is needed)

Sure, if you already have one, you can use that environment. The yaml is only needed to create a new one with all dependencies already installed.

**Comment by maxnoe** [comment]
> But I guess it's cleaner to have a dedicated pyirf env for the editing and testing (that it would be cleaner)

That is the idea, yes.

> but that I could then use my version of the pyirf in other environments by installing using (from the pyirf directory):
```conda env update -f environment.yml``

No, just do the `pip install -e .` in that env. Updating the env using the conda env update command will make your other env equal to what is defined in the yaml, also uninstalling packages you have in that env but that are not present in the yaml. Which is probably not what you want.

**Comment by mdpunch** [comment]
For the failing tests for pyirf/io/tests/test_gadf.py, should I edit the test also? (not sure the problem is with my changes).

**Comment by maxnoe** [comment]
@mdpunch It is another thing where gammpy 0.19 has a different API. Not releated to your changes in `pyirf/gammapy` but just more things that need to be changed to make it work with gammapy 0.19.

**Comment by mdpunch** [comment]
Okay, but then I can look at that too...

**Comment by maxnoe** [comment]
No, we have to fix all incompatibilities with gammapy 0.19 here, since we need to update it all at once to make the tests pass

**Comment by mdpunch** [comment]
But looks like I have to update my local, to take into account the commits you made to master, right? (since I have no tests/io on my local).
I have tried some things, but am worried I am getting into a mess...

File: environment.yml [hunk 1]
```diff
@@ -28,4 +28,5 @@ dependencies:
   - gammapy=0.18
```
maxnoe, line 28: Bump to 0.19 here

environment.yml
```diff
@@ -28,4 +28,5 @@ dependencies:
...
   - pip
   - pip:
+    - gammapy~=0.19.0
```
maxnoe, line 31: Remove here

maxnoe:
@mdpunch I think the remaining issues in the `test_gadf` can be fixed by replacing e.g. `aeff2d.data.data` with `aeff2d.quantity` and similarly for the other IRFs.

pyirf/gammapy.py
```diff
@@ -36,19 +36,17 @@ def create_effective_area_table_2d(
     fov_offset_bins: astropy.units.Quantity[angle]
         Bin edges in the field of view offset.
         For Point-Like IRFs, only giving a single bin is appropriate.
-
```
maxnoe, line 39: These newlines are actually needed. This is what makes the docs build fail.

pyirf/gammapy.py
```diff
@@ -122,18 +120,18 @@ def create_energy_dispersion_2d(
...
```
maxnoe, line 125: And here

maxnoe:
@mdpunch What do you mean with "it worked"? It is not an error that will pop up when running the tests, it creates warnings when building the documentation (and result in broken html in the output).

maxnoe (approved):
Thanks @mdpunch

mdpunch:
> @mdpunch What do you mean with "it worked"? It is not an error that will pop up when running the tests, it creates warnings when building the documentation (and result in broken html in the output)

I mean that when I run ```cd docs && make html SPHINXOPTS="-W --keep-going -n --color -j auto"``` I didn't get any error messages without those blank lines before `Returns`.

I didn't look at any html outputs though ... I got a crash with 
````
[IPKernelApp] ERROR | KeyboardInterrupt caught in kernel.
/scr/punch/miniconda3/envs/pyirf/lib/python3.7/site-packages/sphinx/util/parallel.py:89: RuntimeWarning: coroutine 'ZMQSocketChannel.get_msg' was never awaited
  ret = (errmsg, traceback.format_exc())

Sphinx parallel build error:
nbsphinx.NotebookError: TypeError in notebooks/fact_example.ipynb:
'coroutine' object is not subscriptable
make: *** [Makefile:20: html] Error 2
```

I didn't take out any blank lines, so the error must have been there before ??

But certainly the docstring in the code look nicer with the blank lines!

mdpunch:
> Thanks @mdpunch

Thanks for your patience @maxnoe , I'm still on the learning curve!

PR #165, opened 2021-11-26
Do not require private DL2 event display output anymore for unit tests

maxnoe:
This will enable contributions from forks again.


HealthyPear (approved):

HealthyPear (changes requested):
Sorry I didn't see the CI docs fail: we should install numpydoc

HealthyPear:
> Sorry I didn't see the CI docs fail: we should install numpydoc

Actually it seems all fine: `numpydoc` is one of the docs dependencies and the CI is installing them...

HealthyPear:
Ah I think it's because we are still running the example from `examples/calculate_eventdisplay_irfs.py` in the docs job

maxnoe:
The docs build should be fixed now after rebasing with the current master after the latest PR was merged

HealthyPear:
Either we remove the ED comparison notebook everywhere or we upload a static version

maxnoe:
Done

HealthyPear (approved):

PR #174, opened 2022-01-31
Adapted quantile interpolation

RuneDominik:
This is an alternative interpretation to the approach taken in #163. Instead of following the steps of [this tech report](https://engineering.ucsc.edu/sites/default/files/technical-reports/UCSC-SOE-13-13.pdf) closely, @maxnoe suggested to implement different solutions to the steps summarized in the tech reports Fig. 5.

On the upside, this method should scale easily to nD interpolation variables without using Delaunay simplices. This comes, however, at the price of worse results, at least in the testing case in my comment in #161.

### DIFF
--- a/pyirf/interpolation.py
+++ b/pyirf/interpolation.py
@@ -145,10 +145,10 @@ def norm_pdf(pdf_values):
     """
     norm = np.sum(pdf_values, axis=-1)
 
-    # Ignore errors from rows without errors as they are set to 0 through nan_to_num
-    with np.errstate(invalid="ignore"):
-        normed_pdf_values = pdf_values / norm[..., np.newaxis]
-    return np.nan_to_num(normed_pdf_values)
+    # Norm all binned_pdfs to unity that are not empty
+    normed_pdf_values = np.divide(pdf_values, norm[..., np.newaxis], out=np.zeros_like(pdf_values), where=norm[..., np.newaxis] != 0)
+
+    return normed_pdf_values
 
 
 def interpolate_binned_pdf(edges, binned_pdf, m, mprime, axis, quantile_resolution):
@@ -218,7 +218,7 @@ def interpolate_binned_pdf(edges, binned_pdf, m, mprime, axis, quantile_resoluti
     normed_interpolated_pdfs = norm_pdf(interpolated_pdfs)
 
     # Re-swap axes and set all nans to zero
-    return np.swapaxes(normed_interpolated_pdfs, axis, -1)
+    return np.swapaxes(np.nan_to_num(normed_interpolated_pdfs), axis, -1)
 
 
 @u.quantity_input(effective_area=u.m ** 2)
--- a/pyirf/interpolation.py
+++ b/pyirf/interpolation.py
@@ -25,7 +25,7 @@ def cdf_values(binned_pdf):
     return cdfs
 
 
-def ppf_values(cdfs, bin_mids, quantiles):
+def ppf_values(cdfs, edges, quantiles):
     """
     Compute ppfs from cdfs and interpolate them to the desired interpolation point
 
@@ -34,8 +34,8 @@ def ppf_values(cdfs, bin_mids, quantiles):
     cdfs: numpy.ndarray, shape=(N,...,M)
         Corresponding cdf-values for all quantiles
 
-    bin_mids: numpy.ndarray, shape=(M)
-        M bin-midss for which the cdf-values are known
+    edges: numpy.ndarray, shape=(M+1)
+        Binning of the given binned pdf
 
     quantiles: numpy.ndarray, shape=(L)
         L quantiles for which the ppf_values are known
@@ -53,6 +53,9 @@ def ppf_values(cdfs, bin_mids, quantiles):
         in between filled bins. Both cases would result in division by zero errors when computing
         the interpolation polynom.
         """
+
+        #cdf = np.append(np.array(0), cdf)
+
         # Find last 0 and first 1 entry
         last_0 = np.max(np.arange(len(cdf))[cdf == 0]) if cdf[0] == 0 else 0
         first_1 = np.min(np.arange(len(cdf))[cdf == 1])
@@ -76,6 +79,8 @@ def ppf_values(cdfs, bin_mids, quantiles):
             bounds_error=False,
             fill_value="extrapolate",
         )(quantiles)
+
+    bin_mids = bin_center(edges)
 
     # create ppf values from cdf samples via interpolation of the cdfs
     # return nan for empty pdfs
@@ -198,21 +203,26 @@ def interpolate_binned_pdf(edges, binned_pdf, m, mprime, axis, quantile_resoluti
     # not safely propageted through the interpolation but the last element remains the last element
     binned_pdf = np.swapaxes(binned_pdf, axis, -1)
 
+    # include 0-bin at first position in each pdf to avoid edge-effects where the CDF would otherwise 
+    # start at a value != 0, also extend edges with one bin to the left 
+    fill_zeros = np.zeros(shape=binned_pdf.shape[:-1])[..., np.newaxis]
+    binned_pdf = np.concatenate((fill_zeros, binned_pdf), axis=-1)
+    edges = np.append(edges[0] - np.diff(edges)[0], edges)
+
     # compute quantiles from quantile_resolution
     quantiles = np.linspace(0, 1, int(np.round(1 / quantile_resolution, decimals=0)))
-
-    bin_mids = bin_center(edges)
 
     cdfs = cdf_values(binned_pdf)
 
     # compute ppf values at quantiles, determine quantile step of [1]
-    ppfs = ppf_values(cdfs, bin_mids, quantiles)
+    ppfs = ppf_values(cdfs, edges, quantiles)
 
     # interpolate ppfs to target point, interpolate quantiles step of [1]
     interpolated_ppfs = griddata(m, ppfs, mprime)
 
-    # compute pdf values for all bins, evaluate interpolant PDF values step of [1]
-    interpolated_pdfs = pdf_from_ppf(quantiles, interpolated_ppfs, edges)
+    # compute pdf values for all bins, evaluate interpolant PDF values step of [1], drop the earlier 
+    # introduced extra bin 
+    interpolated_pdfs = pdf_from_ppf(quantiles, interpolated_ppfs, edges)[..., 1:]
 
     # Renormalize pdf to sum of 1
     normed_interpolated_pdfs = norm_pdf(interpolated_pdfs)
@@ -281,13 +291,14 @@ def interpolate_energy_dispersion(
     )
 
 
+@u.quantity_input(psf=u.sr**-1, source_offset_bins=u.deg)
 def interpolate_psf_table(
     psfs,
     grid_points,
     target_point,
     source_offset_bins,
-    cumulative=False,
-    method="linear",
+    axis, 
+    quantile_resolution=1e-3,
 ):
     """
     Takes a grid of PSF tables for a bunch of different parameters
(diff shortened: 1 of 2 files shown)


chaimain:
Is there any update on this or #163?

maxnoe (changes requested):
A general comment, also for the other PR:

Please don't implement the actual interpolation logic inside `interpolate_energy_dispersion`.

Implement functions `interpolate_binned_pdf_<method>(binned_pdfs, edges, grid_points, interpolation_point, **options)` (That should be the required arguments, right?)

This function should then be just applied along the relevant axes in `interpolate_energy_dispersion` and `interpolate_psf` which could have a `method=` argument if we decide to merge both approaches.

We cannot make the assumption of equal distance bins, at least not in general. We could make them in the interpolation method if that is strictly required, but then `interpolate_energy_dispersion` should check for equal-distance logspace binning in migra and then do the interpolation in logspace.

Also please split up the steps inside into single functions. E.g. the resampling of the interpolated pdf back to the original bin edges.

pyirf/interpolation.py
```diff
@@ -54,45 +55,118 @@ def interpolate_effective_area_per_energy_and_fov(
...
+    ] = 0  # 1.1 to correct for numerical uncertainty and interpolation
+    return u.Quantity(aeff_interp, u.m ** 2, copy=False)
 
 
 def interpolate_energy_dispersion(
-    energy_dispersions,
-    grid_points,
-    target_point,
-    method='linear',
+    edges, hists, m, mprime, axis, quantile_resolution=1e-3
```
maxnoe, line 65: hists should be `binned_pdf`, we don't expect a histogram here (as in event counts), correct?

pyirf/interpolation.py
```diff
@@ -54,45 +55,118 @@ def interpolate_effective_area_per_energy_and_fov(
...
-        array of parameters corresponding to energy_dispersions, of shape (n_grid_points, n_interp_dim)
-    target_point: np.ndarray
-        values of parameters for which the interpolation is performed, of shape (n_interp_dim)
-    method: 'linear’, ‘nearest’, ‘cubic’
-        Interpolation method
+    edges: numpy.ndarray, shape=(M+1)
+        Common array of bin-edges (along the abscissal ("x") axis) for the M bins of the input templates
+
+    hists: numpy.ndarray, shape=(N,...,M,...)
+        Array of M bin-heights (along the ordinate ("y") axis) for each of the 2 input templates.
```
maxnoe, line 81: N input templates?

pyirf/interpolation.py
```diff
@@ -54,45 +55,118 @@ def interpolate_effective_area_per_energy_and_fov(
...
+                fill_value=(0, 0),
+            )(bin_mids)
+        ),
+        -1,
+        xyconcat,
+    )
+
+    # Renormalize histogram to a sum of 1
+    norm = np.sum(result, axis=-1)
+    norm = np.repeat(np.expand_dims(norm, -1), result.shape[-1], axis=-1)
```
maxnoe, line 166: Why is repeating / expanding dims necessary here? Should be broadcasting in the division below be enough?

maxnoe:
```
norm = np.sum(result, axis=1)
with np.errstate(invalid="ignore"):
    result /= norm[..., np.newaxis]
result = np.nan_to_num(result)
```

pyirf/interpolation.py
```diff
@@ -2,22 +2,238 @@
...
+        if np.sum(cdf) > 0
+        else np.full_like(quantiles, np.nan),
+        -1,
+        cdfs,
+    )
+    # nD interpolation of ppf values
+    return griddata(m, ppfs, mprime)
+
+
+def reconstruct_pdf_values(quantiles, ppfs, edges):
```
maxnoe, line 100: `pdf_from_ppf`?

pyirf/interpolation.py
```diff
@@ -2,22 +2,238 @@
...
+def norm_pdf(pdf_values):
+    """
+    Normalize binned_pdf to a sum of 1
+    """
+    norm = np.sum(pdf_values, axis=-1)
+
+    # Ignore errors from rows without errors as they are set to 0 through nan_to_num
+    with np.errstate(invalid="ignore"):
+        normed_pdf_values = pdf_values / norm[..., np.newaxis]
+    return np.nan_to_num(normed_pdf_values)
```
maxnoe, line 158: I think the solution before was  a tad more elegant:
```
normed_pdf_values = np.divide(pdf_values, norm, out=np.zeros_like(pdf_values), where=norm != 0)```

Comment (maxnoe) [reply pyirf/interpolation.py:158]: This would also have the advantage of preserving real nans that did not come from dividing through a 0 norm here but were there before

File: pyirf/interpolation.py [hunk 6]
```diff
@@ -2,22 +2,238 @@
...
+           https://engineering.ucsc.edu/sites/default/files/technical-reports/UCSC-SOE-13-13.pdf
+    """
+    # To have the needed axis always at the last index, as the number of indices is
+    # not safely propageted through the interpolation but the last element remains the last element
+    binned_pdf = np.swapaxes(binned_pdf, axis, -1)
+
+    # compute quantiles from quantile_resolution
+    quantiles = np.linspace(0, 1, int(np.round(1 / quantile_resolution, decimals=0)))
+
+    # compute bin-mids from bin-edges
```
maxnoe, line 211: not a helpful comment

pyirf/interpolation.py
```diff
@@ -2,22 +2,238 @@
...
+    bin_mids = bin_center(edges)
+
+    # compute cdf values
+    cdfs = cdf_values(binned_pdf)
+
+    # compute ppf values at interpolation point, determine quantile and interpolate quantiles steps of [1]
+    ppfs = ppf_values(cdfs, bin_mids, m, mprime, quantiles)
```
maxnoe, line 218: With not doing the interpolation inside `ppf_values` above, this would very clearly show the structure of the algorithm here!

maxnoe:
```
ppfs = ppf_values(...)
interpolated_ppfs = griddate(...)
interpolated_pdfs = ppf_to_pdf(...)
...
```

pyirf/interpolation.py
```diff
@@ -2,22 +2,238 @@
...
+
+
+@u.quantity_input(effective_area=u.m ** 2)
 def interpolate_effective_area_per_energy_and_fov(
     effective_area,
     grid_points,
     target_point,
-    min_effective_area=1. * u.Unit('m2'),
-    method='linear',
+    min_effective_area=1.0 * u.Unit("m2"),
```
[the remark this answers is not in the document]
maxnoe: `1 * u.m**2`

pyirf/interpolation.py
```diff
@@ -54,45 +270,22 @@ def interpolate_effective_area_per_energy_and_fov(
     aeff_interp = griddata(grid_points, effective_area, target_point, method=method).T
     # exp it and set to zero too low values
     aeff_interp = np.exp(aeff_interp)
-    aeff_interp[aeff_interp < min_effective_area * 1.1] = 0  # 1.1 to correct for numerical uncertainty and interpolation
-    return u.Quantity(aeff_interp, u.m**2, copy=False)
+    aeff_interp[
+        aeff_interp < min_effective_area * 1.1
+    ] = 0  # 1.1 to correct for numerical uncertainty and interpolation
```
maxnoe, line 275: Black kind of did us a disservice here, I'd prefer:

```
# 1.1 to correct for numerical uncertainty and interpolation
aeff_interp[aeff_interp < min_effective_area * 1.1] = 0 
```

pyirf/interpolation.py
```diff
@@ -132,14 +325,20 @@ def interpolate_psf_table(
     if cumulative:
         psfs = np.cumsum(psfs, axis=3)
 
-    psf_interp = griddata(grid_points, psfs.to_value('sr-1'), target_point, method=method) * u.Unit('sr-1')
+    psf_interp = griddata(
```
maxnoe, line 328: Why not use the new method also for the PSF?

pyirf/interpolation.py
RuneDominik: This results in errors as numpy seems to be unable to broadcast this correctly

pyirf/interpolation.py
```diff
@@ -281,13 +291,14 @@ def interpolate_energy_dispersion(
...
 
+@u.quantity_input(psf=u.sr**-1, source_offset_bins=u.deg)
 def interpolate_psf_table(
     psfs,
     grid_points,
     target_point,
     source_offset_bins,
-    cumulative=False,
-    method="linear",
+    axis,
```
maxnoe, line 300: Why do we need an axis keyword?

maxnoe: Format should be fixed, so axis should be known

pyirf/tests/test_interpolation.py
```diff
@@ -2,18 +2,130 @@
...
+
+
+@pytest.fixture
+def data():
+    """Create common dataset containing binned Gaussians for interpolation testing. Binned PDFs sum to 1."""
+    bin_edges = np.linspace(-5, 30, 101)
+    distributions = [norm(5, 1), norm(10, 2), norm(15, 3)]
+
+    # create binned pdfs by interpolation of bin content
+    norm_1 = distributions[0].cdf(bin_edges[1:]) - distributions[0].cdf(bin_edges[:-1])
```
maxnoe, line 16:
```
norms = np.array([
    np.diff(dist.cdf(bin_edges))
    for dist in distributions
])
```

maxnoe: Also `norms -> binned_pdfs`?

RuneDominik:
The `cdf_values`-function assures that all cdfs max out at 1.

pyirf/interpolation.py
```diff
@@ -2,22 +2,245 @@
...
+    # Unconventional solution to make this usable with np.apply_along_axis for readability
+    # The ppf bin-mids are computed since the pdf-values are derived through derivation
+    # from the ppf-values
+    xyconcat = np.concatenate(
+        (ppfs[..., :-1] + np.diff(ppfs) / 2, pdf_interpolant), axis=-1
+    )
+
+    # Interpolate pdf samples and evaluate at bin edges, weight with the bin_width to estimate
+    # correct bin height via the midpoint rule formulation of the trapezoidal rule
+    pdf_values = np.apply_along_axis(
```
maxnoe, line 128: I assume this was black? This looks horrible...

Maybe better just define a normal function using `def` in here instead of using a lambda?

pyirf/interpolation.py
```diff
@@ -2,22 +2,245 @@
...
+    pdf_values: numpy.ndarray, shape=(1,...,M)
+        Recomputed, binned pdf
+    """
+    # recalculate pdf values through numerical differentiation
+    pdf_interpolant = np.nan_to_num(np.diff(quantiles) / np.diff(ppfs, axis=-1))
+
+    # Unconventional solution to make this usable with np.apply_along_axis for readability
+    # The ppf bin-mids are computed since the pdf-values are derived through derivation
+    # from the ppf-values
+    xyconcat = np.concatenate(
```
maxnoe, line 122: Can't we just call it twice?

pyirf/interpolation.py
```diff
@@ -2,22 +2,245 @@
...
+        (ppfs[..., :-1] + np.diff(ppfs) / 2, pdf_interpolant), axis=-1
```
maxnoe, line 123: isn't this `bin_centers(ppfs)`?

pyirf/interpolation.py
maxnoe: Or find another solution. I really don't like this "clever" solution, especially since the array is split again into the two parts again in the lambda

maxnoe (approved):
Thanks, this looks very good!

PR #191, opened 2022-08-17
Check Table in angular_separation functions

nbiederbeck:
Hi there,

I spent too much time figuring out why my angular resolution was as bad as it was. I found out it was because of how [`astropy.coordinates.angular_separation`](https://docs.astropy.org/en/stable/_modules/astropy/coordinates/angle_utilities.html#angular_separation) behaves when given floats *and* quantities (i.e. one in deg, the other in rad).

`angular_separation` is what is used in our `pyirf.utils.calculate_theta`.

The problem was giving a `Table` instead of a `QTable`, where the columns are no quantities, thus the values get interpreted as floats and thus radians instead of degree (if they were in degree of course).

1) I added our `check_table` function to the functions that use `angular_separation`.
2) I checked for `.unit` being `None`, thus filtering out also the case where a `Table` ist passed (instead of a `QTable`). This fails with some "Column ... has incompatible unit None, expected ...". Here we could decide on checking `isinstance(events, QTable)`.
3) I fixed the very repetitive wording of the wrong unit exception

The thing with these changes is, that incorrect usage before just yields wrong results, where the error (on the user side) is not obvious, after these changes incorrect usage fails.

### DIFF
--- a/pyirf/tests/test_utils.py
+++ b/pyirf/tests/test_utils.py
@@ -1,7 +1,8 @@
-import numpy as np
 import astropy.units as u
-from astropy.table import QTable, Table
+import numpy as np
 import pytest
+from astropy.table import QTable, Table
+
 from pyirf.exceptions import MissingColumns, WrongColumnUnit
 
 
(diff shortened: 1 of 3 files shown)


pyirf/utils.py
```diff
@@ -158,5 +168,5 @@ def check_table(table, required_columns=None, required_units=None):
                 raise MissingColumns(col)
 
             unit = table[col].unit
-            if not expected.is_equivalent(unit):
+            if not unit or not expected.is_equivalent(unit):
```
nbiederbeck, line 171: FYI: this changes the error from
```python
TypeError: None is not a valid Unit
```
to 
```python
WrongColumnUnit: Column "..." has incompatible unit "None", expected "...".
```

RuneDominik, line 171: The `not unit`-case should be tested in `tests/test_utils`., e.g. by passing a `Table` instead of a `QTable`.

RuneDominik (changes requested):

pyirf/utils.py
nbiederbeck: Done

pyirf/tests/test_utils.py
```diff
@@ -49,3 +49,59 @@ def test_check_table():
 
     # m is convertible
     check_table(t, required_units={'bar': u.cm})
+
+    t = Table({'bar': [0, 1, 2]})
+    with pytest.raises(WrongColumnUnit):
```
HealthyPear, line 54: Since you explicitly defined the error message in the exceptions module I think you can be explicit also in its testing,

`with pytest.raises(WrongColumnUnit, match=expected_error_message)`

same for `test_calculate_theta`

HealthyPear (changes requested):

kosack:
I would say if your function expects a QTable later on, you should just explicitly cast the input table to QTable in the function,
```python
def my_func(table):
    qtable = QTable(table)  # then use qtable later in the function   
```
That way you don't have to check for two different ways to access the columns, and then you also don't require QTables to be used as input (which lose access to some column metadata)

pyirf/tests/test_utils.py
nbiederbeck: Done

HealthyPear (approved):

kosack:
What is the status of this quite old PR? Is it still an issue?

pyirf/utils.py
```diff
@@ -147,6 +157,7 @@ def check_table(table, required_columns=None, required_units=None):
         as keys in ``required_units are`` not present in the table.
     WrongColumnUnit: if any column has the wrong unit
     """
+    table = QTable(table)
```
maxnoe, line 160: should pass `copy=False` if we want to take this

PR #197, opened 2022-09-29
Improvements for calculating sensitivity on diffuse gammas

maxnoe:
As one step towards calculating sensitivity in bins of offset, this changes the `estimate_background` method to have an upper and lower limit for the source fov offset instead of just an upper limit.


TarekHC:
I can confirm this PR provides perfect sensitivity matching when using eventDisplay DL2 tables (with cutclass 2). I had to combine another PR (#199 ) to make it work.

TarekHC (dismissed):
We should also update eventDisplay code examples, as they will stop working after this PR.

TarekHC (approved):

PR #199, opened 2022-10-14
Improve under/overflow and nan handling in binning related functions

maxnoe:
Underflow and overflow are now marked with special values (intmin, intmax) that should prevent silent mistakes such as indexing the last entry because the underflow index was -1 before. Using undeflown values to index into a histogram now correctly raises an index error.

The benchmarks / calculate_percentile_cut functions no use the `nan<>` family of functions, so they ignore nan values in their input.

### DIFF
--- a/pyirf/benchmarks/angular_resolution.py
+++ b/pyirf/benchmarks/angular_resolution.py
@@ -1,5 +1,5 @@
 import numpy as np
-from astropy.table import Table
+from astropy.table import QTable
 from scipy.stats import norm
 import astropy.units as u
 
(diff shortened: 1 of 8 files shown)


TarekHC:
I can confirm this PR was working properly once I changed the following:
```python
table = QTable({"values": values, "bin_values": bin_values}, copy=False)
```
to:
```python
table = QTable({"values": values, "bin_values": bin_values}, copy=True)
```
I tested it comparing eventDisplay "cut class 2" events sensitivity vs pyirf sensitivity, showing perfect matching.

maxnoe:
@TarekHC I opened https://github.com/astropy/astropy/issues/13840 and will add a workaround here

JBernete (dismissed):
Every thing looks good to me

maxnoe:
@JBernete I missed a bug (you too ;) ) where I confused quantile and percentile of an untested function.

I fixed it and added a test case. Please review again.

JBernete:
@maxnoe I see, I agree with the change. I'm not still experienced with checks and I see there is a warning that some checks were not succesful. How important is this for the PR?

maxnoe:
@JBernete The docs build fail is crucuial, that needs to pass. The coverage is "nice to have" and should not be too bad.

maxnoe:
The docs build seems to have been a transient connection problem downloading something

JBernete (approved):

PR #200, opened 2022-10-20
Fix unit handling of powerlaw

maxnoe:
This fixes an issue that the astropy exponentiation does not remove the units leading to unexpected results:

```
In [2]: (u.GeV / u.TeV)**-2.67
Out[2]: Unit("TeV(267/100) / GeV(267/100)")
```

### DIFF
--- a/pyirf/spectral.py
+++ b/pyirf/spectral.py
@@ -91,13 +91,17 @@ class PowerLaw:
     )
     def __init__(self, normalization, index, e_ref=1 * u.TeV):
         """Create a new PowerLaw spectrum"""
+        if index > 0:
+            raise ValueError(f'Index must be < 0, got {index}')
+
         self.normalization = normalization
         self.index = index
         self.e_ref = e_ref
 
     @u.quantity_input(energy=u.TeV)
     def __call__(self, energy):
-        return self.normalization * (energy / self.e_ref) ** self.index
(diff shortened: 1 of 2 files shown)


pyirf/spectral.py
```diff
@@ -91,13 +91,17 @@ class PowerLaw:
...
+            raise ValueError(f'Index must be < 0, got {index}')
+
         self.normalization = normalization
         self.index = index
         self.e_ref = e_ref
 
     @u.quantity_input(energy=u.TeV)
     def __call__(self, energy):
-        return self.normalization * (energy / self.e_ref) ** self.index
+        e = (energy / self.e_ref).to_value(u.one)
```
HealthyPear, line 103: I think that the dimensionless unit is now called with `u.dimensionless_unscaled` in astropy, but I see that `u.one` still works and has the same effect, though I can't find it anymore in the documentation.

Perhaps it changed name between v4.0.2 and v5.1.
Anyway, I have no strong feelings, I think they are both readable enough.

HealthyPear (approved):

pyirf/spectral.py
maxnoe: `u.one` exists as long as I can remember and I don't think it's going anywhere...

It's also used in astropy's own code a lot.

PR #202, opened 2022-11-14
Update gammapy requirement to 1.x

maxnoe:


HealthyPear:
I guess this will unblock #170 ?

HealthyPear:
I think you need to update **setup.py**,

extras_require["all"] = list(set(extras_require["tests"] + extras_require["docs"] **+ extras_require["gammapy"]**))

maxnoe:
Yes, that was inconsistent, but the problem was actually that we need to drop python 3.7 support (NEP29, which gammapy also follows)

HealthyPear (approved):
For me it's good.

(I removed the superseded `test (3.7)` CI status check)

PR #203, opened 2022-11-15
Prepare 0.8

maxnoe:


docs/changelog.rst
```diff
@@ -5,36 +5,42 @@ Changelog
...
-.. **Contributors:**
-..
-.. In alphabetical order by last name:
-..
-.. - . . .
-..
-.. Pull Requests
-.. +++++++++++++
-..
-.. - [#XXX] TITLE (AUTHOR)
```
HealthyPear, line 37: Instead of simply removing it, can we save this information  under a new section for developers/maintainers under `/docs/contribute.rst`?

HealthyPear (changes requested):

docs/changelog.rst
maxnoe: I found it easier just to copy an older release and adapt the info....

maxnoe:
Actually, these changelogs at the moment are only duplicating information from the github releases page (which we now nicely automated using release drafter), so It will be better just to link to the release page.

HealthyPear (approved):

PR #204, opened 2022-11-20
optimization of multiplicity, theta and gh cut

maxnoe:


pyirf/cut_optimization.py
```diff
@@ -1,19 +1,205 @@
...
+    background: astropy.table.QTable
+        event list of simulated background events.
+        Required columns are `reco_source_fov_offset`, `reco_energy`,
+        'weight', `gh_score`.
+        No directional (theta) or gamma/hadron cut should already be applied.
+    reco_energy_bins: astropy.units.Quantity[energy]
+        Bins in reconstructed energy to use for sensitivity computation
+    gh_cut_efficiencies: np.ndarray[float, ndim=1]
+        The cut efficiencies to scan for best sensitivity.
+    theta_cuts: astropy.table.QTable
```
kosack, line 51: This doc is wrong: it's  not the cuts being passed here, but the array of values over which the brute-force optimization scan is done, right? Also, for symmetry, shouldn't there be a similar input for multiplicity?

pyirf/cut_optimization.py
```diff
@@ -1,19 +1,205 @@
...
+
+    signal_bin_index, signal_valid = calculate_bin_indices(
+        signal['reco_energy'], reco_energy_bins,
+    )
+    background_bin_index, background_valid = calculate_bin_indices(
+        background['reco_energy'], reco_energy_bins,
+    )
+
+    with tqdm(total=n_cuts, disable=not progress) as bar:
+        for multiplicity_index, multiplicity_cut in enumerate(multiplicity_cuts):
```
kosack, line 89: It might be useful to eventually adapt this to use [scipy.optimize.brute](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brute.html) internally rather than implement the grid-search manually, as that would make it easier to replace it with more efficient optimizers later

pyirf/cuts.py
```diff
@@ -93,6 +107,37 @@ def calculate_percentile_cut(
...
+    Evaluate a binned cut as defined in cut_table on given events.
+
+    Events with bin_values outside the bin edges defined in cut table
+    will be set to False.
+
+    Parameters
+    ----------
+    values: ``~numpy.ndarray`` or ``~astropy.units.Quantity``
+        The values on which the cut should be evaluated
+    cut_table: ``~astropy.table.Table``
```
kosack, line 121: missing doc for bin_index

kosack (changes requested):

pyirf/cut_optimization.py
maxnoe: Using the manual grid search, we can be a bit cleverer than brute force: make the cuts in the outermost loop possible rather than computing all cuts for all computations.

pyirf/cut_optimization.py
```diff
@@ -1,19 +1,205 @@
...
+
+            signal_mask_multiplicity = signal['multiplicity'] >= multiplicity_cut
+            background_mask_multiplicity = background["multiplicity"] >= multiplicity_cut
+
+            gh_cuts = calculate_percentile_cut(
```
maxnoe, line 94: There is a major bug here: cuts are computed per multiplicity cut but not stored that way, the variable is overriden.

That results below in the `best_theta_cuts` and `best_gh_cuts` not actually being the best cuts, but the ones at the same position from the last iteration of the multiplicity loop.

pyirf/cut_optimization.py
```diff
@@ -86,6 +86,8 @@ def optimize_cuts(
     )
 
     with tqdm(total=n_cuts, disable=not progress) as bar:
+        all_the_gh_cuts = []
```
maxnoe, line 89: maybe call this `gh_cut_grid`

`all_the_...` was an example I wrote in chat to clarify the point... not really a good variable name choice I have to admit, same for the theta cuts

kosack (approved):

PR #210, opened 2023-02-07
Rewrite interpolation structure to utilize extendible class structure

RuneDominik:
While the current structure of `pyirf`s interpolation functionalities might suffice with only few interpolation algorithms implemented, extending it with further algorithms (e.g. Moment-Morphing and extrapolation) would result in at least some I/O-Code doubling. This PR refactors the current interpolation algorithms into a class-structure with the upside of having:

- Somewhat defined in- and output so algorithms can simply be exchanged
- Predefined data sanity-checks (e.g. not each interpolation method has to check if the target-point lies within the grid)
- Predefined "interface" to use future extrapolators


pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,235 @@
...
+            # Delaunay.find_simplex() returns -1 for points outside the grids convex hull
+            simplex_ind = self.Triangulation.find_simplex(target_point)
+            return simplex_ind >= 0
+
+    def __call__(self, target_point, extrapolator=None, **kwargs):
+        """Providing a common __call__ interface sanity checking the target point
+
+        Parameters
+        ----------
+            target_point: np.ndarray
```
maxnoe, line 82: Indentation is off here, should start at the beginning

pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,235 @@
...
+                When grid_points is not a np.ndarray
+            TypeError:
+                When grid_point has dtype object
+            ValueError:
+                When there are too few points in grid_points to span a volume
+                in the grid dimension.
+        """
+        if not isinstance(grid_points, np.ndarray):
+            raise TypeError("Input grid_points is not a numpy array.")
+        if grid_points.dtype == "O":
```
kosack, line 35: if you expect the dtype to be, e.g. floats or similar, you might use [np.can_cast](https://numpy.org/doc/stable/reference/generated/numpy.can_cast.html) to check for it to be what is expected, rather than just checking that is not an object.

RuneDominik: This is meant to only prevent inputs like 
```
grid_points = np.array([[1, 2], [1, 2, 3]], dtype=object)
```
to be accepted. Thinking about how often this might occur this actually seems quite defensive, but should not hurt to have it.

kosack: But it just needs to be something numerical for interpolation to work correct? It seems odd to only check for objects, when strings would also be bad (e.g. `np.array([1,2],[1,2,3], dtype='S'` should also fail with a wrong dtype error)

RuneDominik: Good point, added a check and a test

pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,241 @@
...
+                f"To few points for grid dimension, grid-dim is {self.grid_dim},"
+                f" while there are only {self.n_points}. At least {self.grid_dim+1}"
+                f" points needed to span a volume in {self.grid_dim} dimensions."
+            )
+
+        # Build triangulation to check if target is inside of the grid for
+        # more then 1 dimension
+        if self.grid_dim > 1:
+            try:
+                self.Triangulation = Delaunay(self.grid_points)
```
maxnoe, line 63: member variable should be lowercase

pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,241 @@
...
+            except QhullError:
+                raise
+
+    @abstractmethod
+    def interpolate(self, target_point, **kwargs):
+        """Overridable function for the actual interpolation code"""
+        pass
```
maxnoe, line 70: pass not needed here

pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,241 @@
...
+                f" {target_point.shape[1]}."
+            )
+
+        if self._target_in_grid(target_point):
+            return self.interpolate(target_point, **kwargs)
+        elif extrapolator is not None:
+            print(f"Trying to extrapolate for point {target_point}.")
+            try:
+                return extrapolator(target_point, **kwargs)
+            except:
```
maxnoe, line 130: Don't use bare excepts, as these also catch `SystemExit` and `KeyboardInterrupt`. 

Using `except Exception` will catch all other exceptions.

However, here you anyway only reraise, so the try except block should just completely be removed

pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,241 @@
...
+            raise TypeError("Input params is not a numpy array.")
+        elif self.n_points != params.shape[0]:
+            raise ValueError(
+                "Shape missmatch, number of grid_points and rows in params not matching."
+            )
+        else:
+            self.params = params
+
+        if self.params.ndim == 1:
+            self.params = self.params.reshape(*self.params.shape, 1)
```
maxnoe, line 179:
```suggestion
            self.params = self.params[..., np.newaxis]
```

pyirf/interpolators/base_interpolators.py
```diff
@@ -0,0 +1,241 @@
...
+            When number of histograms in bin_contents and points in grid_points
+            is not matching
+
+        Note
+        ----
+            Also calls pyirf.interpolators.BaseInterpolators.__call__
+        """
+        super().__init__(grid_points)
+
+        if not isinstance(bin_edges, np.ndarray):
```
maxnoe, line 223: Why don't we support e.g. quantities here?

pyirf/interpolators/tests/test_base_interpolators.py
```diff
@@ -0,0 +1,184 @@
...
+    bin_content_bad_nbins = np.ones(shape=(len(grid_points), len(bin_edges)))
+
+    with pytest.raises(TypeError):  # Abstract class, cannot be instantiated
+        BinnedInterpolator(grid_points, bin_edges, bin_content_good_shape)
+
+    class DummyBinnedInterpolator(BinnedInterpolator):
+        def interpolate(self, target_point, **kwargs):
+            raise NotImplementedError
+
+    with pytest.raises(TypeError):  # bin_edges are not a np.ndarray
```
maxnoe, line 167: Instead of having a comment with the reason, you can use `match="...."` to actually make sure you get the expected message

pyirf/tests/test_interpolation.py
```diff
@@ -116,7 +9,7 @@ def test_interpolate_effective_area_per_energy_and_fov():
     n_th = 1
     en = np.logspace(-2, 2, n_en)
     # applying a simple sigmoid function
-    aeff0 = 1.0e4 / (1 + 1 / en ** 2) * u.Unit("m2")
+    aeff0 = 1.0e4 / (1 + 1 / en**2) * u.Unit("m2")
```
maxnoe, line 12: u.m**2

maxnoe (changes requested):
Looks mostly good, some comments inline.

Due to the changes, the top-level API (the functions in `interpolate.py`) no seem all untested, which is not good, that should be fixed.

maxnoe (changes requested):
The only thing I don't quite like yet is the structure. Having submodules `interpolation` and `interpolators` right next to each other is confusing.


I'd create a single submodule `interpolation` (directory) and move the `interpolation.py` to something like `interpolation/interpolate_irfs.py` and then import those and the other classes in `interpolation/__init__.py`, that way, the imports also stay backwards compatible with what is currently in `interpolation.py`.

docs/interpolation.rst
```diff
@@ -4,19 +4,18 @@ Interpolation of IRFs
 =====================
 
 This module contains functions to interpolate from a set of IRFs for different
-conditions to a new IRF. While ``pyirf.interpolators`` contains the implementations 
-of interpolation algorithms, ``pyirf.interpolation`` applies them to IRF components.
+conditions to a new IRF. While ``pyirf.interpolation`` contains the implementations
```
maxnoe, line 7: This doesn't make sense anymore since it is the same module.

docs/interpolation.rst
```diff
@@ -4,9 +4,8 @@ Interpolation of IRFs
 =====================
 
 This module contains functions to interpolate from a set of IRFs for different
-conditions to a new IRF. While ``pyirf.interpolation`` contains the implementations 
-of interpolation algorithms as interpolators as well as scripts applying those to 
-IRF components.
+conditions to a new IRF. Implementations of interpolation algorithms exist as interpolator
+objects and are applied by top-level scripts to IRF components.
```
maxnoe, line 8: objects → classes, scripts → functions?

maxnoe (approved):

PR #214, opened 2023-02-23
Use prod5 irf fixture in AEFF interpolation test

RuneDominik:
This PR changes the test for the aeff interpolation to use actual irfs instead of a dummy effective area.


maxnoe:
I actually like the test with specific, artificially created test data better checking a specific result better. 
This is a much more stringent test than what you have now on the prod5b IRFs.

RuneDominik:
Keep both to assert it works on "real" data and give a more understandable test to future developers? While stringent, the old test is also unintuitive in its data creation.

maxnoe (approved):

PR #220, opened 2023-03-16
Update readthedocsconfig

maxnoe:


HealthyPear (approved):
I guess this should fix the build.

PR #221, opened 2023-04-11
Moment morph interpolator

RuneDominik:
This PR introduces another interpolation method (Moment Morphing, see [1]). While performing comparably to the already existing Quantile Interpolation the main points in favor of this method is that it can be can be separated into two steps. First, one has to compute interpolation coefficients and then these are used to morph and interpolate the input pdfs. When changing the coefficients (e.g. changing interpolation to extrapolation or changing from interpolation in a triangle to interpolation in a rectangle) the second part remains unchanged. Consequently, using this for extrapolation is quite simple.

Some performance studies can be found in these notebooks: [PerformanceStudy.zip](https://github.com/cta-observatory/pyirf/files/11200226/PerformanceStudy.zip)

This PR is quite extensive in its additions. Most of these are actually tests.

[1] M. Baak, S. Gadatsch, R. Harrington and W. Verkerke (2015). Interpolation between multi-dimensional histograms using a new non-linear moment morphing method. Nucl. Instrum. Methods Phys. Res. A 771, 39-48. https://doi.org/10.1016/j.nima.2014.10.033

### DIFF
--- a/pyirf/interpolation/moment_morph_interpolator.py
+++ b/pyirf/interpolation/moment_morph_interpolator.py
@@ -1,0 +1,307 @@
+from abc import ABCMeta, abstractmethod
+
+import numpy as np
+
+from .base_interpolators import BinnedInterpolator
+
+__all__ = [
+    "BaseMomentMorphInterpolator",
+    "Base2DTriangularMomentMorphInterpolator",
+    "Base1DMomentMorphInterpolator",
+    "MomentMorphInterpolator",
+]
+
+
+class BaseMomentMorphInterpolator(BinnedInterpolator, metaclass=ABCMeta):
+    def __init__(self, grid_points, bin_edges, bin_contents):
+        super().__init__(grid_points, bin_edges, bin_contents)
+
+    @abstractmethod
+    def get_interpolation_coefficients(self, target_point):
+        """Overridable function creating interpolation coefficients"""
+
+    def _estimate_mean_std(self):
+        """
+        Function to roughly estimate mean and standart deviation from a histogram.
+
+        Returns
+        -------
+        mean: numpy.ndarray, shape=(N)
+            Estimated mean for each input template
+
+        std: numpy.ndarray, shape=(N)
+            Estimated standart deviation for each input template. Set to width/2 if only one bin in
+            the input template is =/= 0
+        """
+        # Create an 2darray where the 1darray mids is repeated n_template times
+        # mids = np.outer(np.ones(self.bin_contents.shape[0]), self.bin_mids)
+        mids = np.broadcast_to(self.bin_mids, self.bin_contents.shape)
+        # Weighted averages to compute mean and std
+        # mean = np.average(mids, weights=self.bin_contents, axis=-1)
+        # std = np.sqrt(np.average((mids.T - mean).T**2, weights=self.bin_contents, axis=-1))
+        mean = np.average(mids, weights=self.bin_contents, axis=-1)
+        std = np.sqrt(
+            np.average(
+                (mids - mean[..., np.newaxis]) ** 2, weights=self.bin_contents, axis=-1
+            )
+        )
+
+        # Set std to 0.5*width for all those templates that have only one bin =/= 0. In those
+        # cases mids-mean = 0 and therefore std = 0. Uses the width of the one bin with
+        # bin_content!=0
+        mask = std == 0
+        if np.any(mask):
+            width = np.diff(self.bin_edges) / 2
+            std[mask] = width[self.bin_contents[mask, :] != 0]
+
+        return mean, std
+
+    def _lookup(self, x):
+        """
+        Function to return the bin-height at a desired point.
+
+        Parameters
+        ----------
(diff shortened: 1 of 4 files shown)


pyirf/interpolation/moment_morph_interpolator.py
```diff
@@ -0,0 +1,442 @@
...
+        j = np.arange(0, self.n_points, 1)
+        m_ij = (self.grid_points - m0) ** j
+
+        # Compute coefficients, eq. (6) from [1]
+        return np.einsum(
+            "...j, ji -> ...i", ((target_point - m0) ** j), np.linalg.inv(m_ij)
+        )
+
+
+class MomentMorphInterpolator(BinnedInterpolator):
```
maxnoe, line 344: I think it's confusing that `MomentMorphInterpolator` is not a child class of `BaseMomentMorph` interpolator. The class hierarchy is a bit unintuitive.

pyirf/interpolation/moment_morph_interpolator.py
```diff
@@ -0,0 +1,442 @@
...
+    def __init__(self, grid_points, bin_edges, bin_contents, axis):
+        """
+        Actual interpolator class utilizing MomentMorphInterpolations.
```
maxnoe, line 347: Remove actual

maxnoe (changes requested):
Without going into more details, the class hierarchy and the intended use is not clear to me here.

maxnoe:
I think what is also needed, is a short, comparative discussion of the different interpolation methods in the docs

RuneDominik:
I guess the main point of confusion here is, that there are two separate structures. There are 
- `BaseMomentMorphInterpolator` which does implement everything needed for Moment Morphing except the interpolation coefficients which are dependent on the dimension of the interpolation parameter space.
- `Base1DMomentMorphInterpolator` and `Base2DTriangularMomentMorphInterpolator`, implementing these coefficients for their respective dimensionality and in the 2D case for a triangular tessellation of the grid (rectangular 2D Moment Morphing would also be possible)

These three classes only operate on one simplex in the respective tessellation, so either a line-segment or a triangle. These can later-on easily reused to provide extrapolation, which is the primary reason why they are separated from the interpolator class that should be used by the user. It should, however, be possible to combine these three classes into one if wanted.

To make this easily usable with a grid of arbitrary size, there is the actual `MomentMorphInterpolator`, the second structure, which takes the full grid, checks its dimensionality, tessellates its, finds the correct simplex for a target-point and then uses the aforementioned classes to actually create a interpolation result.

I hope this cleared things up a bit. I'm open for suggestions, especially for less ambiguous naming schemes.

pyirf/interpolation/moment_morph_interpolator.py
```diff
@@ -0,0 +1,442 @@
...
+        cs = self.get_interpolation_coefficients(target_point)
+
+        # Estimate mean and std for each input template histogram. First adaption needed to extend
+        # the moment morph procedure to histograms
+        mus, sigs = self._estimate_mean_std()
+        cs = cs.reshape(self.n_points, *np.ones(mus.ndim - 1, "int"))
+
+        # Transform mean and std as in eq. (11) and (12) in [1]
+        # cs = np.broadcast_to(cs, mus.shape)
+        mu_strich = np.sum(cs * mus, axis=0)
```
maxnoe, line 162: german?

pyirf/interpolation/moment_morph_interpolator.py
```diff
@@ -0,0 +1,442 @@
...
+        Note
+        ----
+            Also calls BaseMomentMorphInterpolator's __init__.
+        """
+        if grid_points.shape != (3, 2):
+            raise ValueError("This base class can only interpolate in a triangle.")
+
+        super().__init__(grid_points, bin_edges, bin_contents)
+
+    def _target_in_grid(self, target_point):
```
maxnoe, line 226: This is bad: _target_in_grid now "lies" to circumvent a check that is not needed for something that can extrapolate.

maxnoe: If you want to keep the base class structure, I think it would be better to keep the `_target_in_grid` method untouched, since it is well-defined.

Instead, change the check where necessary to something like:

```
if not self.can_extrapolate and not self._target_in_grid(target_point):
    raise ValueError("Cannot extrapolate")
```

pyirf/interpolation/moment_morph_interpolator.py
```diff
@@ -0,0 +1,442 @@
...
+        """
+        super().__init__(grid_points, bin_edges, bin_contents)
+
+    @abstractmethod
+    def get_interpolation_coefficients(self, target_point):
+        """Overridable function creating interpolation coefficients"""
+
+    def _estimate_mean_std(self):
+        """
+        Function to roughly estimate mean and standart deviation from a histogram.
```
maxnoe, line 44: typo standart

RuneDominik:
Superseded by #229

Issue #224, opened 2023-04-19
Allow for event weights when calculating percentile cuts


gschwefer:
When calculating a cut using `cuts.calculate_percentile_cut`, it would be useful to have to the option to associate weight with the `values` argument.

This can be done through an extra argument `weights` to `cuts.calculate_percentile_cut` that defaults to equal weights for all values. Then the `np.nanpercentile` in `cuts.calculate_percentile_cut` would have to be replaced by a function that calculates weighted percentiles.

gschwefer:
If this strategy seems ok I can quickly open a pull request

gschwefer:
Yes, the weighted quantile stuff I have implemented in a custom function similar to the one in stackoverflow. I'll open a pull request

PR #225, opened 2023-04-19
Allow for event weights when calculating percentile cuts

gschwefer:
Pull request implementing issue #224 . Solution as described there, the weighted quantile function is custom.


pyirf/cuts.py
```diff
@@ -44,9 +92,14 @@ def calculate_percentile_cut(
     min_events: int
         Bins with less events than this number are replaced with ``fill_value``
     """
+
+    if weights is None:
```
maxnoe, line 96: I'd prefer it (and it's probably also faster) if the `weights=None` case would use the existing code instead of the `weighted_quantile` method with weights=1.

PR #232, opened 2023-07-03
Add Nearest Neighbor Searcher as Alternative for Inter-/Extrapolation

RuneDominik:
This PR adds an NearestNeighborSearcher Inter-/Extrapolation class for convenience. The name was chosen to reflect that this does not do any Inter-/Extrapolation at all.

### DIFF
--- a/pyirf/interpolation/base_interpolators.py
+++ b/pyirf/interpolation/base_interpolators.py
@@ -2,6 +2,7 @@
 from abc import ABCMeta, abstractmethod
 
 import numpy as np
+import pyirf.interpolation as pyint
 from pyirf.binning import bin_center
 
 __all__ = ["BaseInterpolator", "ParametrizedInterpolator", "DiscretePDFInterpolator"]
@@ -14,7 +15,7 @@ class BaseInterpolator(metaclass=ABCMeta):
     """
 
     def __init__(self, grid_points):
-        """BaseInterpolator  
+        """BaseInterpolator
 
         Parameters
         ----------
@@ -77,6 +78,13 @@ class ParametrizedInterpolator(BaseInterpolator):
         if self.params.ndim == 1:
             self.params = self.params[..., np.newaxis]
 
+    @classmethod
+    def __subclasshook__(cls, subclass):
+        if subclass is pyint.ParametrizedNearestNeighborSearcher:
+            return True
+        else:
+            return NotImplemented
+
 
 class DiscretePDFInterpolator(BaseInterpolator):
     """
@@ -109,3 +117,10 @@ class DiscretePDFInterpolator(BaseInterpolator):
         self.bin_edges = bin_edges
         self.bin_mids = bin_center(self.bin_edges)
         self.bin_contents = bin_contents
+
+    @classmethod
+    def __subclasshook__(cls, subclass):
+        if subclass is pyint.DiscretePDFNearestNeighborSearcher:
+            return True
+        else:
+            return NotImplemented
--- a/pyirf/interpolation/base_interpolators.py
+++ b/pyirf/interpolation/base_interpolators.py
@@ -2,7 +2,6 @@
 from abc import ABCMeta, abstractmethod
 
 import numpy as np
-import pyirf.interpolation as pyint
 from pyirf.binning import bin_center
 
 __all__ = ["BaseInterpolator", "ParametrizedInterpolator", "DiscretePDFInterpolator"]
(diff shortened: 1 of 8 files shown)


pyirf/interpolation/component_estimators.py
```diff
@@ -212,39 +217,61 @@ def __init__(
             grid_points,
         )
 
-        if not isinstance(bin_edges, np.ndarray):
-            raise TypeError("Input bin_edges is not a numpy array.")
-        elif not isinstance(bin_contents, np.ndarray):
+        # Ignore bin_edges if both inter- and extrapolator are NearestNeighborSeacher
```
maxnoe, line 220: mmh... this breaks somehow the separation of concerns... and it's pretty weird to ignore a required, positional argument

pyirf/interpolation/component_estimators.py
```diff
@@ -212,39 +217,61 @@ def __init__(
...
 
         if extrapolator_kwargs is None:
             extrapolator_kwargs = {}
 
-        if not issubclass(interpolator_cls, DiscretePDFInterpolator):
+        if issubclass(interpolator_cls, DiscretePDFInterpolator):
+            self.interpolator = interpolator_cls(
+                grid_points, bin_edges, bin_contents, **interpolator_kwargs
+            )
+        elif issubclass(interpolator_cls, NearestNeighborSearcher):
```
maxnoe, line 259: I'd rather also give the `NearestNeighborSearcher` the bin_edges and ignore them there so that all these estimators have the same api then have this class handle different cases depending on the interpolator.

The argument for this structure was that people can implement their own inter/extra polators, but now, some extrapolators need special treatment here?

maxnoe: But the user *has* to give something to the component estimator, right? It is just ignored when the estimator is the nearest neighbor finder.

RuneDominik: Agreed, although I'd rather not call them "Interpolator" nor would I distinct between Inter- and Extrapolators in this very special case, as this is not really any of these.

pyirf/interpolation/component_estimators.py
```diff
@@ -212,31 +217,35 @@ def __init__(
...
         if interpolator_kwargs is None:
             interpolator_kwargs = {}
 
         if extrapolator_kwargs is None:
             extrapolator_kwargs = {}
 
-        if not issubclass(interpolator_cls, DiscretePDFInterpolator):
+        if not (
+            issubclass(interpolator_cls, DiscretePDFInterpolator)
+            or issubclass(interpolator_cls, DiscretePDFNearestNeighborSearcher)
```
maxnoe, line 244: you can [register](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register) a class as   a virtual subclass of an abstract base class. I think this is appropriate here, the neighbor classes implement the correct API without actually inheriting.

THis would leave the `component_estimator` classes completely unchanged/

pyirf/interpolation/base_interpolators.py
```diff
@@ -109,3 +117,10 @@ def __init__(self, grid_points, bin_edges, bin_contents):
         self.bin_edges = bin_edges
         self.bin_mids = bin_center(self.bin_edges)
         self.bin_contents = bin_contents
+
+    @classmethod
+    def __subclasshook__(cls, subclass):
```
maxnoe, line 122: that's not really what I meant, although it works of course.

Did the documentation I link not work for some reason?

E.g. I expected this here after the class definition of the nearest neighbor classes:
```
DiscretePDFInterpolator.register(DiscretePDFNearestNeighborSearcher)
ParametrizedInterpolator.register(ParametrizedNearestNeighborSearcher)

```

RuneDominik: The link certainly did work, reading into this mechanic I merely thought that having this inside the class definition provides better overview. Is there an advantage in using `.register` over `__subclasshook__` or, directly asked, do you want me to change it to `.register`?

maxnoe: Yes. If this lives inside the base class, users cannot do it for their implementation. The register approach allows anyone to add their own implementation. And then its about showing in our own code how we think it should be done.

maxnoe: Another argument is that a base class shouldn't really have to know anything about its children. If we decide to remove or change `DiscretePDF...` we shouldn't need to modify the parent class.

pyirf/interpolation/base_interpolators.py
```diff
@@ -2,6 +2,7 @@
 from abc import ABCMeta, abstractmethod
 
 import numpy as np
+import pyirf.interpolation as pyint
 from pyirf.binning import bin_center
```
maxnoe, line 6: Use relative imports here

pyirf/interpolation/base_interpolators.py
RuneDominik: makes sense, changed

pyirf/interpolation/base_interpolators.py
RuneDominik: no longer needed when using `.register`

PR #234, opened 2023-07-03
Allow setting which quantile to use for angular resolution

maxnoe:


pyirf/benchmarks/tests/test_angular_resolution.py
```diff
@@ -45,11 +51,22 @@ def test_angular_resolution(unit):
...
+
+    # 2 sigma coverage interval
+    quantile = norm(0, 1).cdf(2) - norm(0, 1).cdf(-2)
+    table = angular_resolution(events, bins, quantile=0.95)
+    ang_res = table['angular_resolution'].to(u.deg)
+
+    assert len(ang_res) == 2
+
+    expected_q = 1 - ((1 - quantile) / 2)
+    print(expected_q)
```
RuneDominik, line 70: Probably superflous print

pyirf/benchmarks/tests/test_angular_resolution.py
```diff
@@ -45,11 +51,22 @@ def test_angular_resolution(unit):
...
+    bins = [1, 10, 100] * u.TeV
 
+    table = angular_resolution(events, bins)
+    ang_res = table['angular_resolution'].to(u.deg)
     assert len(ang_res) == 2
     assert u.isclose(ang_res[0], TRUE_RES_1 * u.deg, rtol=0.05)
     assert u.isclose(ang_res[1], TRUE_RES_2 * u.deg, rtol=0.05)
+
+    # 2 sigma coverage interval
+    quantile = norm(0, 1).cdf(2) - norm(0, 1).cdf(-2)
```
RuneDominik, line 63: Maybe add a utility function for this? The user will need to compute this anyways if they want to compute anything but the 1sigma angular resolution, so we might as well provide it. Its ofc. a one-liner though. Could be reused for the default value in `angular_resolution.py`

maxnoe: common values are stuff like 68%, 90 %, or 95%. 90 is not close to any integer multiple of standard deviations...

This was more to have  a good test case.

pyirf/benchmarks/tests/test_angular_resolution.py
```diff
@@ -45,11 +51,21 @@ def test_angular_resolution(unit):
...
+    table = angular_resolution(events, bins, quantile=0.95)
+    ang_res = table['angular_resolution'].to(u.deg)
+
+    assert len(ang_res) == 2
+
+    expected_q = 1 - ((1 - quantile) / 2)
```
RuneDominik, line 69: unused variable `expected_q`?

PR #239, opened 2023-07-14
Add lower viewcone limit

maxnoe:
To enable computing IRFs for simulations that have also a lower viewcone limit (e.g. LST ring-wobble MCs), add this missing information here.

### DIFF
--- a/pyirf/tests/test_simulations.py
+++ b/pyirf/tests/test_simulations.py
@@ -122,4 +122,16 @@ def test_viewcone_integral():
     expected = _viewcone_pdf_integral(vmin, vmax, 4.5 * u.deg, 5.0 * u.deg)
     assert _viewcone_pdf_integral(vmin, vmax, 4.5 * u.deg, 5.5 * u.deg) == expected
 
-    assert _viewcone_pdf_integral(vmin, vmax, 0 * u.deg, 0.5 * u.deg).ndim == 0
+    # whole region integrates to 1
+    assert _viewcone_pdf_integral(vmin, vmax, 1 * u.deg, 5 * u.deg) == 1.0 * u.one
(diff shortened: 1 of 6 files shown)


pyirf/tests/test_simulations.py
```diff
@@ -100,3 +104,22 @@ def test_integrate_energy_fov_pointlike():
...
+    assert _viewcone_pdf_integral(vmin, vmax, 0 * u.deg, 0.5 * u.deg) == 0 * u.one
+    assert _viewcone_pdf_integral(vmin, vmax, 5.0 * u.deg, 5.5 * u.deg) == 0 * u.one
+
+    # half inside, half outside
+    expected = _viewcone_pdf_integral(vmin, vmax, 1.0 * u.deg, 1.5 * u.deg)
+    assert _viewcone_pdf_integral(vmin, vmax, 0.5 * u.deg, 1.5 * u.deg) == expected
+    expected = _viewcone_pdf_integral(vmin, vmax, 4.5 * u.deg, 5.0 * u.deg)
+    assert _viewcone_pdf_integral(vmin, vmax, 4.5 * u.deg, 5.5 * u.deg) == expected
+
+    assert _viewcone_pdf_integral(vmin, vmax, 0 * u.deg, 0.5 * u.deg).ndim == 0
```
RuneDominik, line 125: This test mostly targets edge cases but not really the math inside the `_viewcone_pdf_integral`. Is there any set-up where the analytic solution to the integration is easily reproducible or even written in a paper? At least I'm struggling why the integration results in the implemented equations, so a reference would be helpful.

maxnoe: CORSIKA simulates the particles inside the viewcone as uniform per solid angle, which is the following PDF (copied from my phd thesis):

uniform_solid_angle

maxnoe: Added a test for "in-range" values

RuneDominik (approved):

PR #240, opened 2023-07-14
Fix nearest simplex finding

### LINKED ISSUE #238: Nearest Simplex finding in ParametrizedNearestSimplexExtrapolator not working
Apparently, `scipy.spatial.Delaunay.plane_distance` does not work in all cases to find the nearest simplex to some point outside the grid's convex hull. For non-rectangular grids this does somewhat work
non_rect_scipy
but it fails completely for rectangular grids
rect_scipy

With some assumptions the problem can be solved geometrically to behave as expected:
non_rect_geom
rect_geom

I'll create a PR implementing the solution used to create above plots, some of these functions could even be reused for later extrapolation methods (namely blending over visible edges). I'd like, however, some advise where to put the functions used for this as they are quite general. The first place I have in mind would ofc be just as free functions inside `pyirf.interpolation.nearest_simplex_extrapolator` as they are first needed there, but since they are so general and can partially be reused later, maybe we want something as `pyirf.interpolation.utils` @maxnoe?

As of now, the `ParametrizedNearestSimplexExtrapolator` should not be used. As the same approach was used in #237 this issue also blocks this PR.

RuneDominik:
With this PR, the nearest simplex is computed geometrically instead of using `scipy.spatial.Delaunay.plane_distance`. I've put the functions for this into `pyirf.interpolation.utils`. This was just a first thought, if they are better suited to lay inside `pyirf.interpolation.nearest_simplex_extrapolator` I'll happily change this. Some of these functions can be reused later for an extrapolator that blends over visible edges.

Fixes #238

### DIFF
--- a/pyirf/interpolation/utils.py
+++ b/pyirf/interpolation/utils.py
@@ -1,0 +1,164 @@
+import numpy as np
+
+
+def plumb_point_dist(line, target):
+    """
+    Compute minimal distance between target and line under the constrain, that it has
+    to lay between the points building the line and not on the extention of it.
+
+    Parameters
+    ----------
+    line: np.ndarray, shape=(2, M)
+        Array of two points spanning a line segment. Might be in two or three dims M.
+    target: np.ndarray, shape=(M)
+        Target point, of which the minimal distance to line segement is needed
+
+    Returns
+    -------
+    d_min: float
+        Minimal distance to line segement between points in line
+    """
+    A = line[0]
+    B = line[1]
+    P = target
+
+    # Costruct the footpoint/plumb point of the target projected onto
+    # both  lines F1 = OA + r1*AB and F2 = OB + r1*BA
+    F1 = A + (B - A) * np.dot(P - A, B - A) / np.dot(B - A, B - A)
+    F2 = B + (A - B) * np.dot(P - B, A - B) / np.dot(A - B, A - B)
+
+    # Find, at which parameter value r1/r2 the plumb point lies on line F1/F2
+    if B[0] - A[0] == 0:
+        r1 = (F1[1] - A[1]) / (B[1] - A[1])
+    else:
+        r1 = (F1[0] - A[0]) / (B[0] - A[0])
+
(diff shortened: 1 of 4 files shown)


pyirf/interpolation/utils.py
```diff
@@ -0,0 +1,164 @@
+import numpy as np
+
+
+def plumb_point_dist(line, target):
+    """
+    Compute minimal distance between target and line under the constrain, that it has
```
maxnoe, line 6: constraint

pyirf/interpolation/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    to lay between the points building the line and not on the extention of it.
```
maxnoe, line 7: extension

pyirf/interpolation/utils.py
```diff
@@ -0,0 +1,164 @@
+import numpy as np
+
+
+def plumb_point_dist(line, target):
```
maxnoe, line 4: any references for the formulas here?

RuneDominik: Sure, I've found a wiki entry, hope that suffices. You can find them below, links were also added as comments to the code 
https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Another_formula

maxnoe: > This derivation also requires that the line be not vertical or horizontal. 

Do we handle that case?

RuneDominik: This caveat is, as I see it, only for the analytical deviation, not for the vector ones I used. I messed up the link in the comment above (sorry for that :sweat_smile:), it should redirect to the section after the one it does link to (https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Vector_formulation). The links in the code-comments however are correct. To check things, I've added additional test cases and these cases are as far as the tests go handled correctly.

maxnoe (approved):

PR #241, opened 2023-07-14
Remove nonsensical condition from relative_sensitivity

maxnoe:
The function required that the relative sensitsivity (to the assumed spectrum) must be <= 1, which actually has no special meaning at all, it just means that a source brighter than the  reference spectrum is needed.

This prevented computing the sensitivity of the last two highest energy bins for North Alpha layout compared to event display.

### DIFF
--- a/docs/changes/241.bugfix.rst
+++ b/docs/changes/241.bugfix.rst
@@ -1,0 +1,4 @@
+Remove condition that relative sensitivity must be > 1.
+This condition was added by error and resulted in returning
+nan if the flux needed to fulfill the conditions is larger than
+the reference flux used to weight the events.
--- a/pyirf/sensitivity.py
+++ b/pyirf/sensitivity.py
@@ -67,10 +67,6 @@ def _relative_sensitivity(
         )
         return np.nan
 
-    # less than min_sigma
-    if relative_flux > 1:
-        return np.nan
-
     # scale to achieved flux level
     n_signal = n_signal * relative_flux
     min_excess = min_excess_over_background * n_background
--- a/pyirf/tests/test_sensitivity.py
+++ b/pyirf/tests/test_sensitivity.py
@@ -51,12 +51,13 @@ def test_calculate_sensitivity():
 
     sensitivity = calculate_sensitivity(signal_hist, bg_hist, alpha=0.2)
 
-    # too small sensitivity
+    # sensitivity smaller than 1
     signal_hist['n_weighted'] = [10, 10]
     sensitivity = calculate_sensitivity(signal_hist, bg_hist, alpha=0.2)
 
     assert len(sensitivity) == len(signal_hist)
-    assert np.all(np.isnan(sensitivity['relative_sensitivity']))
+    np.testing.assert_almost_equal(5.0, sensitivity['significance'])
+    np.testing.assert_array_less(1.0, sensitivity['relative_sensitivity'])
 
     # not above 5 percent of remaining background
     signal_hist['n_weighted'] = 699
@@ -74,6 +75,12 @@ def test_calculate_sensitivity():
     # we scale up the signal until we meet the requirement, so
     # we must have more than 5 sigma
     assert np.all(sensitivity['significance'] >= 5)
+
+    # No background at all, li&ma not applicable => event count requirement
+    signal_hist['n_weighted'] = [5, 5]
+    bg_hist['n_weighted'] = [0, 0]
+    sensitivity = calculate_sensitivity(signal_hist, bg_hist, alpha=0.2)
+    np.testing.assert_equal(sensitivity['relative_sensitivity'], 2)
 
 
 def test_estimate_background():


rlopezcoto:
and this was affecting the  two highest energy bins (because the sensitivity was >100%) but no low energy ones?
PS: Tests need to be modified as well

maxnoe:
> and this was affecting the two highest energy bins (because the sensitivity was >100%) but no low energy ones?

Yes, exactly.

maxnoe:
Here is the official prod5 50h sensitivity with Crab Nebula flux levels, only the last (here three) bins are above 1 CU

prod5_sensitivity

rlopezcoto:
ok, thanks! I thought that low energy sensitivity was also overcoming the Crab Nebula flux (maybe in the Southern Array with the alpha configuration?)

maxnoe:
> ok, thanks! I thought that low energy sensitivity was also overcoming the Crab Nebula flux (maybe in the Southern Array with the alpha configuration?)

Without LSTs probably, yes, I didn't check.

RuneDominik (approved):

mdpunch:
> Here is the official prod5 50h sensitivity with Crab Nebula flux levels, only the last (here three) bins are above 1 CU
> 
> prod5_sensitivity

Minor comment @maxnoe @rlopezcoto : this is (presumably) the 50h sensitivity to an $E^{-2}$ spectrum, whereas the Crab is more like $E^{-2.7}$ so a direct comparison between the Crab spectrum and the sensitivity just gives an idea of detectability in the bins.  (I just realized this myself a few days ago, so made a pull request on the Gammapy `SensitivityEstimator` doc to make this clearer).  
There are some CTA papers with the sensitivity for a Crab Spectrum, but from 2015, so probably a bit outdated, e.g. https://arxiv.org/pdf/1509.01943.pdf "Real-Time Analysis sensitivity evaluation of the Cherenkov Telescope Array" which has "The MC simulation assumes a source with a power law spectrum and a Crab-like index located at the center of the field of view and observed at a zenith angle of 20 degrees."

maxnoe:
> Minor comment @maxnoe @rlopezcoto : this is (presumably) the 50h sensitivity to an $E^{-2}$ spectrum, whereas the Crab is more like $E^{-2.7}$ so a direct comparison between the Crab spectrum and the sensitivity just gives an idea of detectability in the bins.

No, the simulated events are weighted to the Crab Nebula flux (per the CTAC document prepared by @GernotMaier and @moralejo  to the Crab spectrum as published by HEGRA), which is:

$$
2.83e-11 \mathrm{TeV}^{-1}\,\mathrm{cm}^{-2}\,\mathrm{s)^{-1} \left(\frac{E}{1\,\mathrm{TeV}}\right)^(-2.62)
$$

However, this is affects only the distribution of energies inside a bin, so I expect the differences to be minor between different fluxes the events are weighted to.

mdpunch:
> > Minor comment @maxnoe @rlopezcoto : this is (presumably) the 50h sensitivity to an E−2 spectrum, whereas the Crab is more like E−2.7 so a direct comparison between the Crab spectrum and the sensitivity just gives an idea of detectability in the bins.
> 
> No, the simulated events are weighted to the Crab Nebula flux (per the CTAC document prepared by @GernotMaier and @moralejo to the Crab spectrum as published by HEGRA), which is:
> 
> 2.83⋅10−11TeV−1,cm−2,s−1(E1,TeV)(−2.62)
> 
> However, this is affects only the distribution of energies inside a bin, so I expect the differences to be minor between different fluxes the events are weighted to.

Oh, interesting, thanks!

For the CTA sensitivity given on the official website https://www.cta-observatory.org/science/ctao-performance/#1472563157332-1ef9e83d-426c there is no description of the spectrum for which the spectrum is calculated, so that information is missing there.  So I suppose I should make a "pull request" on the CTAO info page!

For the zenodo links https://zenodo.org/record/7298569 for the current prod5 simulations, and https://zenodo.org/record/5163273 to previous baseline prod3 simulation, this information is indeed there: "A power-law gamma-ray spectrum with photon index 2.62 was assumed in the calculations."

But as I mentioned `Gammapy` assumes an index of 2 by default.   So many people could fall into the trap.

Finally, I and others had thought, like you, that the change in index for the sensitivity would only move events around within bins.  So much so that we found it hard to believe the effect when changing the spectral model by default used in Gammapy's `SensitivityEstimator`!  But I checked this a bit brutally with my pre-Gammapy sensitivity code (by pruning the gammas as a function of energy to change the index), and I confirm that the effect is striking, especially for softer sources with an index >3 (not to mention MAGIC's GRB with an index of 5.43).

**Conclusion:** Nothing, just for info, but I may push the webmaster of CTAO to include full information on the webpage.

GernotMaier:
Main reason why no spectrum is mentioned is that we are showing differential sensitivities. 5 bins per log decade are probably enough that the effect of the assumed spectral index can be neglected (is well below other uncertainties).

mdpunch:
> Main reason why no spectrum is mentioned is that we are showing differential sensitivities. 5 bins per log decade are probably enough that the effect of the assumed spectral index can be neglected (is well below other uncertainties).

Hi Gernot @GernotMaier ... except that I have seen that it can't with the ALTO WCD array.  I will see if I can do the same demonstration with the CTA IRFs (which maybe indeed isn't so affected, since the energy resolution is better than WCDs)... but I am sometimes surprised...

mdpunch:
So, indeed this is not a violent for CTA:
image
vs.
image

Still, it would be nice to get people into the habit of giving the info, since it will have a bigger effect for HAWC, LHASSO, SWGO, ALTO...

mdpunch:
Superimposed:
image

PR #242, opened 2023-07-18
Update .github workflows

RuneDominik:
This updates the workflows to the newest version and adds a test setup for python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


maxnoe (approved):

maxnoe:
Thanks!

As per NEP-29, we can also drop 3.8 now:

> On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)

RuneDominik:
> Thanks!
> 
> As per NEP-29, we can also drop 3.8 now:
> 
> > On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)

If we want this I'll make this a new PR, arguably we could wait for `ctapipe` to make this move as `pyirf` likely will become a dependency there sometime and we may want to aim for consistency throughout `cta-observatory` packages. However, even them requiring `python >= 3.8` and us `python >= 3.9` is ofc. solvable.

maxnoe:
The PR for this in ctapipe is already open. But that's kind of besides the point. The consistency here is via NEP 29, as soon as we make a new non-bugfix release, we drop versions accordingly.

PR #250, opened 2023-08-02
Normalize edisp to integral of 1, not sum of 1

maxnoe:
@RuneDominik does this have also implications on the interpolation code?

### DIFF
--- a/pyirf/irf/energy_dispersion.py
+++ b/pyirf/irf/energy_dispersion.py
@@ -10,9 +10,10 @@ __all__ = [
 ]
 
 
-def _normalize_hist(hist):
+def _normalize_hist(hist, migration_bins):
     # make sure we do not mutate the input array
     hist = hist.copy()
+    bin_width = np.diff(migration_bins)
 
     # calculate number of events along the N_MIGRA axis to get events
     # per energy per fov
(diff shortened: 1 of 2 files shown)


RuneDominik:
Yes, it does. At the moment `MomentMoprhInterpolator`, `QuantileInterpolator` and `MomentMorphNearestSimplexExtrapolator` also norm to a sum of one. As this however only is a global factor, we could just nomalize by integration in `EnergyDispersionEstimator` after computing the interpolant, right? 

I'm however not sure how this effects the performance of the algorithms. All EDisps I've used so far for testing summed to one as they were computed with pyirf.

pyirf/interpolation/base_extrapolators.py
```diff
@@ -83,7 +84,7 @@ class DiscretePDFExtrapolator(BaseExtrapolator):
     Derived from pyirf.interpolation.BaseExtrapolator
     """
 
-    def __init__(self, grid_points, bin_edges, bin_contents):
+    def __init__(self, grid_points, bin_edges, binned_pdf, normalization=PDFNormalization.AREA):
```
RuneDominik, line 87: Add `normalization` to docstring

pyirf/interpolation/nearest_simplex_extrapolator.py
```diff
@@ -122,7 +124,7 @@ def extrapolate(self, target_point):
 
 
 class MomentMorphNearestSimplexExtrapolator(DiscretePDFExtrapolator):
-    def __init__(self, grid_points, bin_edges, bin_contents):
+    def __init__(self, grid_points, bin_edges, binned_pdf, normalization=PDFNormalization.AREA):
```
RuneDominik, line 127: Add `normalization` to docstring

pyirf/interpolation/tests/__init__.py
RuneDominik, line 1: Is there a reason for this empty file (i guess there is, I simply do not know it :sweat_smile: )

RuneDominik (changes requested):
Besides some minor docstring updates and a missing test I did not find any major things here.

pyirf/interpolation/tests/__init__.py
maxnoe: To have it automatically included in the distribution, only python packages are and for that you need the `__init__.py`.

maxnoe:
@RuneDominik done

RuneDominik (approved):

PR #251, opened 2023-08-21
Start working on fixing interpolation for fixed edisp

maxnoe:
@RuneDominik This is a PR targeting #250 with changes related to the interpolation code.

I started this during ICRC and will have a look now again, I don't remember quite how far I got, please have a look and maybe just push here any changes you might want to include to make it work

### DIFF
--- a/pyirf/interpolation/moment_morph_interpolator.py
+++ b/pyirf/interpolation/moment_morph_interpolator.py
@@ -46,10 +46,10 @@ def _estimate_mean_std(bin_edges, binned_pdf, normalization):
     # binned_pdf!=0
     mask = std == 0
     if np.any(mask):
-        # Create array of bin_widths and clone for each template
-        width = np.diff(bin_edges) / 2
-        width = np.repeat(width[np.newaxis, :], binned_pdf.shape[0], axis=0)
-        std[mask] = width[binned_pdf[mask, :] != 0]
+        width = np.diff(bin_edges)
+        # std of a uniform distribution inside the bin
+        uniform_std = np.broadcast_to(np.sqrt(1/12) * width, binned_pdf[mask].shape)
+        std[mask] = uniform_std[binned_pdf[mask, :] != 0]
 
     return mean, std
 
--- a/pyirf/interpolation/moment_morph_interpolator.py
+++ b/pyirf/interpolation/moment_morph_interpolator.py
@@ -2,14 +2,14 @@ import numpy as np
 from pyirf.binning import bin_center, calculate_bin_indices
 from scipy.spatial import Delaunay
 
-from .base_interpolators import DiscretePDFInterpolator, PDFNormalization
+from .base_interpolators import DiscretePDFInterpolator, PDFNormalization, get_bin_width
 
 __all__ = [
     "MomentMorphInterpolator",
 ]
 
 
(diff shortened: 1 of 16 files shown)


pyirf/interpolation/tests/test_moment_morph_interpolator.py
```diff
@@ -29,14 +29,14 @@ def bins():
 def simple_1D_data(bins):
     grid = np.array([[20], [40]])
     target = np.array([30])
-    bin_contents = np.array([binned_normal_pdf([x, 0], bins) for x in grid])
+    binned_pdf = np.array([binned_normal_pdf([x, 0], bins) for x in grid])
```
RuneDominik, line 32: `binned_normal_pdf` produces test-cases where the pdf sums to one, not integrates to one. As these cases here are all with a linear binning this global factor should not change anything, however we should probably change the test case to better reflect the use-case

pyirf/interpolation/tests/test_nearest_simplex_extrapolator.py
```diff
@@ -138,7 +138,7 @@ def test_MomentMorphNearestSimplexExtrapolator_1DGrid(bins):
     grid_points = np.array([[20], [30], [40]])
 
     # Create template histograms
-    bin_contents = np.array(
+    binned_pdf = np.array(
         [
             [
                 [binned_normal_pdf([x, 0], bins), binned_normal_pdf([x + 1, 0], bins)],
```
RuneDominik, line 144: Same as above, `binned_normal_pdf` sums to 1, not integrates to 1.

pyirf/interpolation/moment_morph_interpolator.py
```diff
@@ -29,35 +29,40 @@ def _estimate_mean_std(bin_edges, bin_contents):
...
     )
 
     # Set std to 0.5*width for all those templates that have only one bin =/= 0. In those
     # cases mids-mean = 0 and therefore std = 0. Uses the width of the one bin with
-    # bin_content!=0
+    # binned_pdf!=0
     mask = std == 0
     if np.any(mask):
         # Create array of bin_widths and clone for each template
         width = np.diff(bin_edges) / 2
```
maxnoe, line 50: @RuneDominik Two things here: 

First, width should be taking normalization into account now, right? 

Second: why is width `diff / 2`? Shouldn't it just be `diff`?

RuneDominik: As this only applies to cases where only one bin has any entry != 0 we should not need the normalization, as the bin-height in this case is irrelevant. 

`diff/2` was chosen instead of `diff` to mimic a Gaussian `N(bin_mid, diff/2)` and thus taking the bin to represent the +-1sigma interval. Arguably, this results in 32% of the probability is outside the bin, even though the data does not support this. This might be a problem for logarithmic binning. It works fine for linear binning, as the dirac-delta test-cases show.

maxnoe: or do `with = np.diff(bin_edges)` and then `std = 0.5 * width`.

One could also just assume a uniform distribution inside the bin, in which case the std would be:
```
np.sqrt(1 / 12 * width**2)
```

RuneDominik (approved):

maxnoe:
Going on in #250

PR #264, opened 2023-09-26
Ignore warnings of  possibly NaN bins in sensitivity

Tobychev:
By putting the multiplications with `rel_sens` into a `errstate` context manager that ignores warnings from the invalid floating point operations the NaNs in  `rel_sens` cause.


Tobychev:
@maxnoe I have no idea why, but I think something has broken the doctests.

maxnoe:
Seems to just be a download failure, I restarted the docs build

maxnoe (approved):

Tobychev:
The problem seems to be pretty enduring.

maxnoe:
Should be fixed now... let's try again

PR #266, opened 2023-10-05
Some clarification of what functions do

Tobychev:
After struggling to understand what these functions do, I reworked the doc strings to be a bit more clear, added missing return value annotations.

### DIFF
--- a/pyirf/binning.py
+++ b/pyirf/binning.py
@@ -129,8 +129,9 @@ def create_bins_per_decade(e_min, e_max, bins_per_decade=5):
 
 def calculate_bin_indices(data, bins):
     """
-    Calculate bin indices for given data array and bins.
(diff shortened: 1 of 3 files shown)


pyirf/binning.py
```diff
@@ -152,6 +153,9 @@ def calculate_bin_indices(data, bins):
         Indices of the histogram bin the values in data belong to.
         Under- and overflown values will have values of `UNDERFLOW_INDEX`
         and `OVERFLOW_INDEX` respectively.
+
+    valid: np.ndarray[bool]
+        Boolean mask indicated if a given value fell into the over- or underflow bins.
```
RuneDominik, line 158: I'd rephrase slightly, as `valid` indicates those values that are neither under- nor overflow. Maybe

Boolean mask indicating those values as valid that did not fell into the over- or underflow bins.

RuneDominik (changes requested):

RuneDominik (approved):

Issue #267, opened 2023-10-17
energy_bias_resolution_from_energy_dispersion still assumes wrong normalization of EDISP


Tobychev:
Plotting the bias+1 from `energy_bias_resolution_from_energy_dispersion` on top of the energy bias matrix gives the following result: 
![Screenshot of a comment on a GitHub issue showing an image, added in the Markdown, of an Octocat smiling and raising a tentacle.](https://filesender.renater.fr/download.php?token=b654321c-faf3-4cd8-99e1-4d53335036ab&files_ids=30321071)

where the blue line and the red line are both estimates of the median in each column of the matrix. The red curve produced using the numpy percentile function using the 'median_unbiased' method (the recommended method in the paper numpy references but for some reason not the default value).

Tobychev:
Fixed by #268

PR #268, opened 2023-10-19
Fix energy_bias_resolution_from_energy_dispersion

maxnoe:
This function still assumed the wrong normalization of the energy dispersion fixed in pyirf 0.10.


Tobychev (approved):

maxnoe:
@Tobychev the mistake was obvious (the function was overlooked in the update), but could you anyway maybe check with your plot in #267 ?

Tobychev:
In case there is nothing in an energy bin `energy_bias_resolution_from_energy_dispersion` will return the upper limit of the dispersion axis. It is probably better to return NaN

Tobychev (dismissed):

Tobychev (approved):

PR #269, opened 2023-11-27
Correct typos in docstrings

morcuended:


RuneDominik:
Oh, that's more typos than I would have thought, thanks for going over them :sweat_smile: 

Tests are failing since the typos were seemingly carried over to test the exceptions being raised correctly.

morcuended:
> Oh, that's more typos than I would have thought, thanks for going over them 😅
> 
> Tests are failing since the typos were seemingly carried over to test the exceptions being raised correctly.

Let me have a look

maxnoe (approved):

morcuended:
I found also that in PSF and EDISP estimators' docstrings, was specified that they default to `GridDataInterpolator` instead of `QuantileInterpolator`. Also noticed that arguments in those classes are not properly included in the documentation since they are written after their `__init__` methods instead of the docstring of the classes themselves.

RuneDominik (approved):

Issue #272, opened 2023-12-06
energy_dispersion_to_migration doesn't produce a matrix as if starting directly from the events


HealthyPear:
When calling `energy_dispersion_to_migration` on a normalized energy dispersion histogram (right) I get a different
result than building the migration matrix directly from the data (center)

the right plot is a crosscheck with gammapy to_edisp_kernel, but in that case, one needs to choose a FoV offset angle (in this case I put 0 deg), but the first 2 plots have all offsets.

![image](https://github.com/cta-observatory/pyirf/assets/17836610/191a1c7a-d841-4ff4-b4ae-f0ac66f44c67)

One culprit seems to be at least the fact that this function was not updated after #268 

I suggest also improving the current units test to compare the 2 histograms.

PR #273, opened 2023-12-12
Update energy_dispersion_to_migration to account for fix in energy dispersion normalization

HealthyPear:
This attempts to update the function that produces the energy migration matrix from the energy dispersion.

After #250 this function was not updated which resulted in the wrong matrix when compared with the one computed starting directly from the events (see #272)

I think I got a working result, but it might still be wrong - also for reasons related to my use case I didn't test this using different binning, so I am re-sampling the same thing to comparing with the original matrix (left panel)

image

I should also update the unit test in order to compare the counts, but since it comes from a re-sampling I am not sure how much big the relative difference should be.

### DIFF
--- a/pyirf/irf/energy_dispersion.py
+++ b/pyirf/irf/energy_dispersion.py
@@ -6,7 +6,8 @@ from ..binning import resample_histogram1d
 
 __all__ = [
     "energy_dispersion",
-    "energy_dispersion_to_migration"
+    "energy_migration_matrix",
+    "energy_dispersion_to_migration",
 ]
 
 
@@ -81,6 +82,56 @@ def energy_dispersion(
     energy_dispersion = _normalize_hist(energy_dispersion, migration_bins)
 
     return energy_dispersion
+
+
+@u.quantity_input(true_energy_bins=u.TeV, reco_energy_bins=u.TeV, fov_offset_bins=u.deg)
+def energy_migration_matrix(
+    events, true_energy_bins, reco_energy_bins, fov_offset_bins
+):
+    """Compute the energy migration matrix directly from the events.
+
+    Parameters
+    ----------
+    events : `~astropy.table.QTable`
+        Table of the DL2 events.
+        Required columns: ``reco_energy``, ``true_energy``, ``true_source_fov_offset``.
+    true_energy_bins : `~astropy.units.Quantity`
+        Bin edges in true energy.
+    reco_energy_bins : `~astropy.units.Quantity`
+        Bin edges in reconstructed energy.
+
+    Returns
+    -------
+    matrix : array-like
+        Migration matrix as probabilities along the true
+        energy acis with shape
+        (n_true_energy_bins, n_reco_energy_bins, n_fov_offset_bins)
+        containing energies in TeV.
+    """
+
+    hist, _ = np.histogramdd(
+        np.column_stack(
+            [
+                events["true_energy"].to_value(u.TeV),
+                events["reco_energy"].to_value(u.TeV),
+                events["true_source_fov_offset"].to_value(u.deg),
+            ]
+        ),
+        bins=[
+            true_energy_bins.to_value(u.TeV),
+            reco_energy_bins.to_value(u.TeV),
+            fov_offset_bins.to_value(u.deg),
+        ],
+    )
+
+    with np.errstate(invalid="ignore"):
+        hist /= hist.sum(axis=1)[:, np.newaxis, :]
+        # the nans come from the fact that the sum along the reconstructed energy axis
+        # might sometimes be 0 when there are no events in that given true energy bin
+        # and fov offset bin
+        hist[np.isnan(hist)] = 0
+
+    return hist
 
 
 def energy_dispersion_to_migration(
--- a/pyirf/irf/energy_dispersion.py
+++ b/pyirf/irf/energy_dispersion.py
@@ -29,7 +29,10 @@ def _normalize_hist(hist, migration_bins):
 
 
 def energy_dispersion(
-    selected_events, true_energy_bins, fov_offset_bins, migration_bins,
+    selected_events,
+    true_energy_bins,
+    fov_offset_bins,
+    migration_bins,
 ):
     """
     Calculate energy dispersion for the given DL2 event list.
@@ -88,8 +91,8 @@ def energy_dispersion_to_migration(
     new_reco_energy_edges,
 ):
     """
-    Construct a energy migration matrix from a energy
-    dispersion matrix.
+    Construct a energy migration matrix from an energy dispersion matrix.
+
     Depending on the new energy ranges, the sum over the first axis
     can be smaller than 1.
     The new true energy bins need to be a subset of the old range,
@@ -110,21 +113,25 @@ def energy_dispersion_to_migration(
     new_reco_energy_edges: astropy.units.Quantity[energy]
         Reco energy edges matching the second dimension of the output
 
-    Returns:
-    --------
+    Returns
+    -------
     migration_matrix: numpy.ndarray
         Three-dimensional energy migration matrix. The third dimension
         equals the fov offset dimension of the energy dispersion matrix.
     """
+    migration_matrix = np.zeros(
+        (
+            len(new_true_energy_edges) - 1,
+            len(new_reco_energy_edges) - 1,
+            dispersion_matrix.shape[2],
+        )
+    )
 
-    migration_matrix = np.zeros((
-        len(new_true_energy_edges) - 1,
-        len(new_reco_energy_edges) - 1,
-        dispersion_matrix.shape[2],
-    ))
+    migra_width = np.diff(disp_migration_edges)
+    probability = dispersion_matrix * migra_width[np.newaxis, :, np.newaxis]
 
     true_energy_interpolation = resample_histogram1d(
-        dispersion_matrix,
+        probability,
         disp_true_energy_edges,
         new_true_energy_edges,
         axis=0,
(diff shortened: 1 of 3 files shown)


pyirf/irf/energy_dispersion.py
```diff
@@ -110,49 +110,55 @@ def energy_dispersion_to_migration(
...
+            dispersion_matrix.shape[2],
+        )
+    )
 
     true_energy_interpolation = resample_histogram1d(
         dispersion_matrix,
         disp_true_energy_edges,
         new_true_energy_edges,
         axis=0,
+        normalize=True
```
maxnoe, line 132: I think you should transform the dispersion_matrix from a pdf to the probabilities before calling `resample_histogram1d` and then the rest of the code can stay the same.

I.e.:
```
resample_histogram1d(
    dispersion_matrix / np.diff(disp_migration_edges)[np.newaxis, :, np.newaxis],
...
)

**Comment by HealthyPear** [comment]
For reference, this is the code I am using to plot the data,

```python
def plot_migration_matrix(true_quantity, reco_quantity, data, quantity_name=None, ax=None, logx=True, logy=True, vmin=0):

    ax = plt.gca() if ax is None else ax

    X, Y = true_quantity, reco_quantity
    if len(data.shape)==3:
        Z = np.sum(data, axis=2).T
    else:
        Z = data.T

    matrix = ax.pcolormesh(X, Y, Z, vmin=vmin)
    fig=plt.gcf()
    fig.colorbar(matrix, ax=ax, label='Normalized number of events')

    if logx:
        ax.set_xscale("log")
    if logy:
        ax.set_yscale("log")
    quantity_name = "" if quantity_name is None else quantity_name
    ax.set_xlabel(f"True {quantity_name} [ {true_quantity.unit} ]")
    ax.set_ylabel(f"Reco {quantity_name} [ {reco_quantity.unit} ]")

    return ax
```

calling it like this,

`plot_migration_matrix(true_energy_bins, reco_energy_bins, migration_matrix, quantity_name="energy")`

pyirf/irf/energy_dispersion.py
HealthyPear: but you still need to normalize the final migration matrix
if you do that + what you propose you obtain the same result I show here

(also sorry that `normalize=True` part is some debugging code I added, I'll remove it)

HealthyPear: I squashed the removal of the debugging code and addition of the news fragment, you should re-pull if you pulled before.

maxnoe:
Tests are fixed in main, please rebase

pyirf/irf/energy_dispersion.py
maxnoe: > but you still need to normalize the final migration matrix

No, why? The interpolation should produce the correct normalization.

HealthyPear: oh wait I think this is because in my code the original matrix was normalized so the sum is 1, not the PDF

in the case of the integral = 1 the peaks can still be greater than 1 so I guess its OK but I guess it's better to add to the test the check that the area is 1

HealthyPear:
I noticed we don't have a function that allows to create the energy migration matrix directly from the events.

I think it would be nice to have it also as a crosscheck for the tests.

pyirf/irf/energy_dispersion.py
HealthyPear: However if one plots the same matrix using gammapy from a kernel (provided we account for the fact that in that case you have to choose an FoV offset) one gets a z-axis which goes up to 1 so it should be normalized with the sum=1

HealthyPear: I think what I did is correct though: the migration matrix should be normalized as having probabilities along the e_reco axis and I checked that the sum along that axis is 1 for each e_true

if I do not normalize the final matrix after dividing by the migration bin widths this is not true anymore

maxnoe: If you handle the input EDISP correctly, not renormalization should be required.

In the first comment in this thread I posted how I think it should be done. Could you please try?

HealthyPear:
I added a function to compute the migration matrix directly from the events with its unit test.

Now the final touch would be to test the two function against each other, but I am afraid that the resampling makes this complex in terms of the tolerance required.

pyirf/irf/energy_dispersion.py
```diff
@@ -155,4 +206,13 @@ def energy_dispersion_to_migration(
 
         migration_matrix[idx, :, :] = y
 
+    with np.errstate(invalid="ignore"):
+        migration_matrix = (
+            migration_matrix
+            / migration_matrix.sum(axis=2).sum(axis=1)[:, np.newaxis, np.newaxis]
```
maxnoe, line 212: Why is this needed? This looks wrong. Summing over the offset axis is definitely not right here.

pyirf/irf/energy_dispersion.py
```diff
@@ -110,40 +163,38 @@ def energy_dispersion_to_migration(
...
+        (
+            len(new_true_energy_edges) - 1,
+            len(new_reco_energy_edges) - 1,
+            dispersion_matrix.shape[2],
+        )
+    )
 
     true_energy_interpolation = resample_histogram1d(
-        dispersion_matrix,
+        dispersion_matrix / np.diff(disp_migration_edges)[np.newaxis, :, np.newaxis],
```
maxnoe, line 181: You need to integrate! So multiplying with the bin_width, not divide

pyirf/irf/energy_dispersion.py
HealthyPear: but you want the probability to be independent of the offset the migration matrix is really a 2D object

pyirf/irf/energy_dispersion.py
HealthyPear: right - this makes it wrong though (right is the migration matrix from the events)

image

pyirf/irf/energy_dispersion.py
HealthyPear: yeah I did that and it didn't work (also it's wrong, should be a multiplication)

pyirf/irf/energy_dispersion.py
HealthyPear: This instead is what happens if you multiply by the bin width instead of dividing and do not renormalize the final matrix

image

HealthyPear: this is instead with the current version of this branch

image

maxnoe: With only the one-line fix I proposed above and no other changes:
```
❯ git diff
diff --git a/pyirf/irf/energy_dispersion.py b/pyirf/irf/energy_dispersion.py
index de631e4..19e9be0 100644
--- a/pyirf/irf/energy_dispersion.py
+++ b/pyirf/irf/energy_dispersion.py
@@ -123,8 +123,10 @@ def energy_dispersion_to_migration(
         dispersion_matrix.shape[2],
     ))
 
+    migra_width = np.diff(disp_migration_edges)
+    probability = dispersion_matrix * migra_width[np.newaxis, :, np.newaxis]
     true_energy_interpolation = resample_histogram1d(
-        dispersion_matrix,
+        probability,
         disp_true_energy_edges,
         new_true_energy_edges,
         axis=0,
```

I get this on main:
edisp_main


and this with the fix:
edisp_fix

As is to be expected by how we changed the definition of the EDISP.

maxnoe: Both are actually not that good... maybe it's the interpolation. I will check how that compares to gammapy's edisp kernel method

maxnoe: gammapy_comparison

maxnoe: To be honest, I lean towards just removing this method and recommending gammapy for users interested in this.

HealthyPear:
> Are the differences in the top two figures understood? 

Please don't worry about these figures - this is SWGO data and I still cannot make this function work with that dataset for some reason. I am reverting to compute the energy migration matrix directly from the events with just one bin in FoV offset because that shows the correct matrix for me.

Together with Max, we ran it on some dummy data with ideal distributions and these issues do not arise in that case.

[edisp_to_matrix.ipynb.zip](https://github.com/cta-observatory/pyirf/files/13672230/edisp_to_matrix.ipynb.zip)

maxnoe:
> Also shouldn't the integral along the x-axis in this case be 1.0, not the peak? It's a probability distribution.

For the migration matrix, the sum should be one, you don't want a PDF but a transfer matrix.

pyirf/irf/energy_dispersion.py
```diff
@@ -80,6 +84,56 @@ def energy_dispersion(
...
+    true_energy_bins : `~astropy.units.Quantity`
+        Bin edges in true energy.
+    reco_energy_bins : `~astropy.units.Quantity`
+        Bin edges in reconstructed energy.
+
+    Returns
+    -------
+    matrix : array-like
+        Migration matrix as probabilities along the true
+        energy acis with shape
```
maxnoe, line 107: typo: acis -> axis

pyirf/irf/energy_dispersion.py
```diff
@@ -80,6 +84,56 @@ def energy_dispersion(
...
+        Required columns: ``reco_energy``, ``true_energy``, ``true_source_fov_offset``.
+    true_energy_bins : `~astropy.units.Quantity`
+        Bin edges in true energy.
+    reco_energy_bins : `~astropy.units.Quantity`
+        Bin edges in reconstructed energy.
+
+    Returns
+    -------
+    matrix : array-like
+        Migration matrix as probabilities along the true
```
maxnoe, line 106: The probabilities are along the reco energy axis. The migration matrix maps from true energy to the reco energies.

pyirf/irf/energy_dispersion.py
HealthyPear: fixed

pyirf/irf/energy_dispersion.py
HealthyPear: fixed

maxnoe (approved):

PR #277, opened 2024-01-18
Fix docs build

maxnoe:


maxnoe:
@RuneDominik this is the fix for the docs  failure in your PR, approve here?

PR #278, opened 2024-01-19
Update table based on DAMPE results

VictorBarbosaMartins:
Hello. 

I updated the table with the DAMPE (still preliminary) results from 2021: https://inspirehep.net/files/f08b6bbb9cade07c5c11b8d3f16e9493

In the paper, the data are given in `m^-2 sr^-1 s^-1 GeV ^1.7 ` but I converted to `m^-2 sr^-1 s^-1 GeV ^-1` as the old data (see comparison plot below):

image

If you check both publications, the new results cover now up to 10^5 GeV.


maxnoe:
Hi @VictorBarbosaMartins 

I'd prefer not to update the table, since it was used and is based on a specific publcation, but rather to add new tables for the new references.

If I understood correctly, these are also separate tables for Protons and Helium, so it would be nice to have them, people can just add the resulting spectra if they want to have the combined flux.

maxnoe:
Ah sorry, this is indeed the same P +He spectrum, just ICRC 2019 vs. 2021. Still, I think we should have both tables and maybe call them accordingly.

maxnoe:
I thought there were publications by know giving the individual fluxes.

VictorBarbosaMartins:
Hi @maxnoe. I returned the file to its previous state and created a new one with the 2021 data points as suggested. I think it would have been better to link the publication to a specific version of the package via e.g. a zenodo doi instead of not updating the data. But I am also not used to this package and do not know how important this actually is.

kosack (approved):

PR #276, opened 2024-01-16
Add basic facilities for 3D background irfs

Tobychev:
Add functions for making a 3D background using square bins, this requires that the event tables have the two columns `reco_fov_lat`  and `reco_fov_lon`.

PR #281, opened 2024-02-15
Effective Area 3D

luca-dib:
Adds two methods of calculating the effective area in bins of energy and two spacial dimensions. These methods differ in the coordinate frame used for the spacial dimensions.

- `calculate_effective_area_3d_polar` uses a polar coordinate system with FOV pointing position offset and position angle coordinates
- `calculate_effective_area_3d_nominal` uses a quasi-cartesian coordinate system with FOV longitude and latitude coordinates centred on the pointing position 

Also adds respective functions to calculate the number of showers expected in the 3D bins from simulation info to the SimulatedEventsInfo class (`calculate_n_showers_3d_polar`, `calculate_n_showers_3d_nominal`), as well as a utility function to calculate the position angle w.r.t. the pointing from an event table (`calculate_source_fov_position_angle`).

### DIFF
--- a/pyirf/coordinates.py
+++ b/pyirf/coordinates.py
@@ -2,8 +2,8 @@ import astropy.units as u
 from astropy.coordinates import SkyCoord, SkyOffsetFrame, angular_separation, position_angle
 
 __all__ = [
-    'fov_coords_lon_lat',
-    'fov_coords_theta_phi'
+    'gadf_fov_coords_lon_lat',
+    'gadf_fov_coords_theta_phi',
 ]
 
 
--- a/pyirf/coordinates.py
+++ b/pyirf/coordinates.py
@@ -1,0 +1,62 @@
+import astropy.units as u
+from astropy.coordinates import SkyCoord, SkyOffsetFrame, angular_separation, position_angle
+
+__all__ = [
+    'fov_coords_lon_lat',
+    'fov_coords_theta_phi'
+]
+
+
+def fov_coords_lon_lat(lon, lat, pointing_lon, pointing_lat):
+    """Transform sky coordinates to field-of-view longitude-latitude coordinates.
+
+    Parameters
+    ----------
+    lon, lat : `~astropy.units.Quantity`
+        Sky coordinate to be transformed.
+    pointing_lon, pointing_lat : `~astropy.units.Quantity`
+        Coordinate specifying the pointing position.
+        (i.e. the center of the field of view.)
+
+    Returns
+    -------
+    lon_t, lat_t : `~astropy.units.Quantity`
+        Transformed field-of-view coordinate.
+    """
+    # Create a frame that is centered on the pointing position
+    center = SkyCoord(pointing_lon, pointing_lat)
+    fov_frame = SkyOffsetFrame(origin=center)
+
+    # Define coordinate to be transformed.
+    target_sky = SkyCoord(lon, lat)
+
+    # Transform into FoV-system
+    target_fov = target_sky.transform_to(fov_frame)
+
+    # Switch sign of longitude angle since this axis is
+    # reversed in our definition of the FoV-system
+    return -target_fov.lon, target_fov.lat
+
+
+def fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
+    """Transform sky coordinates to field-of-view theta-phi coordinates.
+
+    Parameters
+    ----------
+    az, alt : `~astropy.units.Quantity`
+        Sky coordinate to be transformed.
+    pointing_az, pointing_alt : `~astropy.units.Quantity`
+        Coordinate specifying the pointing position.
+        (i.e. the center of the field of view.)
+
+    Returns
+    -------
+    lon_t, lat_t : `~astropy.units.Quantity`
+        Transformed field-of-view coordinate.
+    """
+
+    theta = angular_separation(pointing_az, pointing_alt, az, alt)
+    
+    phi = position_angle(pointing_az, pointing_alt, az, alt)
+    
+    return theta.to(u.deg), (-phi).wrap_at(360 * u.deg).to(u.deg)
--- a/pyirf/coordinates.py
+++ b/pyirf/coordinates.py
@@ -7,8 +7,12 @@ __all__ = [
 ]
 
 
-def fov_coords_lon_lat(lon, lat, pointing_lon, pointing_lat):
-    """Transform sky coordinates to field-of-view longitude-latitude coordinates.
+def gadf_fov_coords_lon_lat(lon, lat, pointing_lon, pointing_lat):
+    """Transform sky coordinates to field-of-view longitude-latitude coordinates in accordance with
+    the definitions laid out by the Gamma Astro Data Format.
+
+    GADF documentation here:
+    https://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html
 
     Parameters
     ----------
@@ -38,8 +42,12 @@ def fov_coords_lon_lat(lon, lat, pointing_lon, pointing_lat):
     return -target_fov.lon, target_fov.lat
 
 
-def fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
-    """Transform sky coordinates to field-of-view theta-phi coordinates.
+def gadf_fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
+    """Transform sky coordinates to field-of-view theta-phi coordinates in accordance with
+    the definitions laid out by the Gamma Astro Data Format.
+
+    GADF documentation here:
+    https://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html
 
     Parameters
     ----------
@@ -51,12 +59,14 @@ def fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
 
     Returns
     -------
-    lon_t, lat_t : `~astropy.units.Quantity`
+    theta, phi : `~astropy.units.Quantity`
         Transformed field-of-view coordinate.
     """
 
     theta = angular_separation(pointing_az, pointing_alt, az, alt)
-    
+
+    # astropy defines the position angle as increasing towards east of north
     phi = position_angle(pointing_az, pointing_alt, az, alt)
-    
+
+    # GADF defines FOV PHI opposite to the position angle way so the sign is switched
     return theta.to(u.deg), (-phi).wrap_at(360 * u.deg).to(u.deg)
--- a/pyirf/coordinates.py
+++ b/pyirf/coordinates.py
@@ -42,7 +42,7 @@ def gadf_fov_coords_lon_lat(lon, lat, pointing_lon, pointing_lat):
     return -target_fov.lon, target_fov.lat
 
 
-def gadf_fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
+def gadf_fov_coords_theta_phi(lon, lat, pointing_lon, pointing_lat):
     """Transform sky coordinates to field-of-view theta-phi coordinates in accordance with
     the definitions laid out by the Gamma Astro Data Format.
 
@@ -51,9 +51,9 @@ def gadf_fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
 
     Parameters
     ----------
-    az, alt : `~astropy.units.Quantity`
+    lon, lat : `~astropy.units.Quantity`
         Sky coordinate to be transformed.
-    pointing_az, pointing_alt : `~astropy.units.Quantity`
+    pointing_lon, pointing_lat : `~astropy.units.Quantity`
         Coordinate specifying the pointing position.
         (i.e. the center of the field of view.)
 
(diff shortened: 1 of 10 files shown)


pyirf/irf/effective_area.py
```diff
@@ -7,6 +7,8 @@
     "effective_area",
     "effective_area_per_energy",
     "effective_area_per_energy_and_fov",
+    "effective_area_3d_polar",
+    "effective_area_3d_nominal",
```
maxnoe, line 11: nominal doesn't really tell much... latlon?

pyirf/irf/effective_area.py
```diff
@@ -85,3 +87,69 @@ def effective_area_per_energy_and_fov(
...
+        The field of view longitude bin edges in which to calculate effective area.
+    fov_latitude_bins: astropy.units.Quantity[angle]
+        The field of view latitude bin edges in which to calculate effective area.
+    """
+    area = np.pi * simulation_info.max_impact ** 2
+    
+    hist_simulated = simulation_info.calculate_n_showers_3d_nominal(energy_bins, fov_longitude_bins, fov_latitude_bins)
+    
+    hist_selected, _ = np.histogramdd(
+        np.array([selected_events['true_energy'].to_value(u.TeV), selected_events['true_source_fov_lon'].value, selected_events['true_source_fov_lat'].value]).T,
```
maxnoe, line 151: `array(...).T` is `column_stack`, right?

maxnoe, line 151: Don't use `.value`, always use `to_value(unit)`.

pyirf/simulations.py
```diff
@@ -168,6 +169,81 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+        n_showers: numpy.ndarray(ndim=3)
+            The expected number of events inside each of the
+            ``energy_bins``, ``fov_offset_bins`` and ``fov_position_angle_bins``.
+            Dimension (n_energy_bins, n_fov_offset_bins, n_fov_position_angle_bins)
+            This is a floating point number.
+            The actual numbers will follow a poissionian distribution around this
+            expected value.
+        """
+        e_fov_offset_integral = self.calculate_n_showers_per_energy_and_fov(energy_bins, fov_offset_bins)
+        position_angle_integral = np.ones(len(fov_position_angle_bins) - 1) * self.calculate_n_showers_per_fov(np.linspace(self.viewcone_min, self.viewcone_max, 2)) / (len(fov_position_angle_bins) - 1)
```
maxnoe, line 202: `np.ones(shape) * something` -> `np.full(shape, something)`

pyirf/simulations.py
```diff
@@ -168,6 +169,115 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+            This is a floating point number.
+            The actual numbers will follow a poissionian distribution around this
+            expected value.
+        """
+        e_fov_offset_integral = self.calculate_n_showers_per_energy_and_fov(
+            energy_bins, fov_offset_bins
+        )
+        position_angle_integral = np.full(
+            len(fov_position_angle_bins) - 1,
+            self.calculate_n_showers_per_fov(
```
maxnoe, line 209: Isn't this just the sum of `e_fov_offset_integal` over the energy axis?

pyirf/simulations.py
```diff
@@ -168,6 +169,115 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+            This is a floating point number.
+            The actual numbers will follow a poissionian distribution around this
+            expected value.
+        """
+        fov_bin_midpoints_lon, fov_bin_midpoints_lat = np.meshgrid(
+            (fov_longitude_bins[:-1] + fov_longitude_bins[1:]) / 2,
+            (fov_latitude_bins[:-1] + fov_latitude_bins[1:]) / 2,
+        )
+        mask_outside_fov = np.logical_or(
+            np.sqrt(fov_bin_midpoints_lon**2 + fov_bin_midpoints_lat**2)
```
maxnoe, line 258: use `angular_separation` instead of small angle approximation

pyirf/simulations.py
```diff
@@ -168,6 +169,115 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+            > self.viewcone_max,
+            np.sqrt(fov_bin_midpoints_lon**2 + fov_bin_midpoints_lat**2)
```
maxnoe, line 260: don't compute the same things twice

pyirf/simulations.py
```diff
@@ -168,6 +169,115 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+        n_showers: numpy.ndarray(ndim=3)
+            The expected number of events inside each of the
+            ``energy_bins``, ``fov_longitude_bins`` and ``fov_latitude_bins``.
+            Dimension (n_energy_bins, n_fov_longitude_bins, n_fov_latitude_bins)
+            This is a floating point number.
+            The actual numbers will follow a poissionian distribution around this
+            expected value.
+        """
+        fov_bin_midpoints_lon, fov_bin_midpoints_lat = np.meshgrid(
+            (fov_longitude_bins[:-1] + fov_longitude_bins[1:]) / 2,
```
maxnoe, line 254: There is a `bin_center` function in `pyirf.binning`

pyirf/utils.py
```diff
@@ -88,6 +90,34 @@ def calculate_source_fov_offset(events, prefix="true"):
...
+        source fov position_angle.
+
+    Returns
+    -------
+    phi: astropy.units.Quantity
+        Position angle of the true positions relative to the pointing positions
+        in the sky.
+    """
+    phi = position_angle(
+        events["pointing_az"],
```
maxnoe, line 112: I am not sure this definition matches the defintion in GADF given here:
https://gamma-astro-data-formats.readthedocs.io/en/v0.3/general/coordinates.html#field-of-view

maxnoe: Please add tests for that

pyirf/irf/effective_area.py
```diff
@@ -85,3 +87,99 @@ def effective_area_per_energy_and_fov(
...
+        - `true_source_fov_offset`
+        - `true_source_fov_position_angle`
+    simulation_info: pyirf.simulations.SimulatedEventsInfo
+        The overall statistics of the simulated events
+    true_energy_bins: astropy.units.Quantity[energy]
+        The true energy bin edges in which to calculate effective area.
+    fov_offset_bins: astropy.units.Quantity[angle]
+        The field of view radial bin edges in which to calculate effective area.
+    fov_position_angle_bins: astropy.units.Quantity[radian]
+        The field of view azimuthal bin edges in which to calculate effective area.
```
kosack, line 116: not sure if this could lead to confusion, but a 2D _azimuthal_ coordinate and _position angle_ are not quite the same: the polar azimuth starts on the positive X-axis and goes counter-clockwise, while position angle starts on the positive Y-axis and goes clockwise, so there is a 90 deg phase shift.  Perhaps just to avoid this, say "Field-of-view azimuthal (from Y-axis, toward positive x) bin edges...".      Probably this should be shown as a figure somewhere in the general PyIRF documentation...

maxnoe: This is going into the same direction as the comment above about being clear how we define these coordinate systems and adding tests for the cases spelled out by GADF.

One confusing thing here is that the position angle is defined relative to the original coordinate system axes (AltAz or RaDec) but the FoV lon has the opposite direction to the original lon. Which I think makes the polar angle go counter-clockwise when looking at in a common plot of field of view coordinates where the cartesian like axes are FoV lon/lat...

pyirf/simulations.py
```diff
@@ -168,6 +169,115 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+            (fov_latitude_bins[:-1] + fov_latitude_bins[1:]) / 2,
+        )
+        mask_outside_fov = np.logical_or(
+            np.sqrt(fov_bin_midpoints_lon**2 + fov_bin_midpoints_lat**2)
+            > self.viewcone_max,
+            np.sqrt(fov_bin_midpoints_lon**2 + fov_bin_midpoints_lat**2)
+            < self.viewcone_min,
+        )
+        A_bins = (
```
maxnoe, line 263: Please implement a new function in `util` to compute the solid angle for these rectangular bins and use it here.

We will need it also for #276

pyirf/irf/effective_area.py
luca-dib: If I understand it correctly the GADF FOV position angle definition is exactly the same as the usual definition of the azimuthal angle in polar coordinates, right? (Angle increasing counter-clockwise)
So following this definition there shouldn't _technically_ be a difference between the two, although I understand that just calling it the azimuthal angle here might still cause confusion.

Since we are following the GADF conventions anyways, would it be agreeable to just change "field of view azimuthal bin edges" to "field of view position angle bin edges" and maybe showing a figure somewhere to clarify as suggested?

pyirf/coordinates.py
```diff
@@ -0,0 +1,62 @@
+import astropy.units as u
+from astropy.coordinates import SkyCoord, SkyOffsetFrame, angular_separation, position_angle
+
+__all__ = [
+    'fov_coords_lon_lat',
+    'fov_coords_theta_phi'
+]
+
+
+def fov_coords_lon_lat(lon, lat, pointing_lon, pointing_lat):
```
maxnoe, line 10: I'd put a `gadf_` in front here, other standards might use different definitions of these coordinates. Also add links to the corresponding gadf documentation.

pyirf/coordinates.py
```diff
@@ -0,0 +1,62 @@
...
+    -------
+    lon_t, lat_t : `~astropy.units.Quantity`
+        Transformed field-of-view coordinate.
+    """
+
+    theta = angular_separation(pointing_az, pointing_alt, az, alt)
+    
+    phi = position_angle(pointing_az, pointing_alt, az, alt)
+    
+    return theta.to(u.deg), (-phi).wrap_at(360 * u.deg).to(u.deg)
```
maxnoe, line 62: The `(-phi).wrap_at()` part deserves a comment

pyirf/irf/effective_area.py
```diff
@@ -85,3 +87,104 @@ def effective_area_per_energy_and_fov(
...
+    Parameters
+    ----------
+    selected_events: astropy.table.QTable
+        DL2 events table, required columns for this function:
+        - `true_energy`
+        - `true_source_fov_lon`
+        - `true_source_fov_lat`
+    simulation_info: pyirf.simulations.SimulatedEventsInfo
+        The overall statistics of the simulated events
+    true_energy_bins: astropy.units.Quantity[energy]
```
maxnoe, line 162: mismatch of keyword name in docstring and function definition

maxnoe: Should always use `true_energy_bins`

pyirf/simulations.py
```diff
@@ -168,6 +172,107 @@ def calculate_n_showers_per_energy_and_fov(self, energy_bins, fov_bins):
...
+            bin_grid_lat[1:,1:],
+        )
+        viewcone_area = cone_solid_angle(self.viewcone_max) - cone_solid_angle(self.viewcone_min)
+
+        shower_density = self.n_showers / viewcone_area
+
+        fov_integral = shower_density * bin_area
+        e_integral = self.calculate_n_showers_per_energy(energy_bins)
+
+        fov_integral = fov_mask * fov_integral
```
maxnoe, line 272: Is fov_mask boolean? No, right? It contains a float computed from the subpixel computation, correct?

I wouldn't call it mask then, mask is conventionally used for boolean indexing.

pyirf/coordinates.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Transform into FoV-system
+    target_fov = target_sky.transform_to(fov_frame)
+
+    # Switch sign of longitude angle since this axis is
+    # reversed in our definition of the FoV-system
+    return -target_fov.lon, target_fov.lat
+
+
+def gadf_fov_coords_theta_phi(az, alt, pointing_az, pointing_alt):
```
maxnoe, line 45: This should probably also be called `lon` / `lat` as it supports both I think.

pyirf/simulations.py
```diff
@@ -224,3 +329,69 @@ def _viewcone_pdf_integral(viewcone_min, viewcone_max, fov_low, fov_high):
     if scalar:
         return np.squeeze(integral)
     return integral
+
+def _fov_lonlat_grid_overlap(self, bin_edges_lon, bin_edges_lat, radius, subpixels=20):
```
maxnoe, line 333: Please add a docstring here although this is a private method

maxnoe: Also: do I see it correctly that this function does not properly handle `viewcone_min > 0`?

luca-dib: Yes you're right. That is of course unintentional, will fix asap. Out of curiosity: when and why is this ever the case?

maxnoe: LST simulates diffuse gammas on a ring for use with the standard wobble observations.

maxnoe (approved):

kosack (approved):

PR #282, opened 2024-03-21
Improve Handling of Fill-Values in RAD_MAX Estimator

RuneDominik:
In the current version, the RAD_MAX estimator does not handle fill-values and thus produces results that inter-/extrapolate between actual values and fill-values, yielding results that do not resemble the expected behavior in edge-cases (e.g. RAD_MAX values sharply rising after some energy, see Fig. below). This PR introduces some form of combinatoric edge-case handling in 1 and 2D grids.

[rad_max_cuts_new.pdf](https://github.com/cta-observatory/pyirf/files/14694256/rad_max_cuts_new.pdf)

### DIFF
--- a/docs/changes/282.feature.rst
+++ b/docs/changes/282.feature.rst
@@ -1,0 +1,1 @@
+Add basic combinatoric fill-value handling for RAD_MAX estimation.
--- a/pyirf/interpolation/component_estimators.py
+++ b/pyirf/interpolation/component_estimators.py
@@ -541,6 +541,12 @@ class RadMaxEstimator(ParametrizedComponentEstimator):
         else:
             self.fill_val = fill_value
 
+        # Raise error if fill-values should be handled in >=3 dims
+        if self.fill_val and self.grid_dim >= 3:
+            raise ValueError(
+                "Fill-value handling only supported in up to two grid dimensions."
+            )
+
         # If fill-values should be handled in 2D, construct a trinangulation
         # to later determine in which simplex the target values is
         if self.fill_val and (self.grid_dim == 2):
@@ -566,19 +572,15 @@ class RadMaxEstimator(ParametrizedComponentEstimator):
         # First, construct estimation without handling fill-values
         full_estimation = super().__call__(target_point)
         # Safeguard against extreme extrapolation cases
-        full_estimation[full_estimation < 0] = 0
+        np.clip(full_estimation, 0, None, out=full_estimation)
 
         # Early exit if fill_values should not be handled
         if not self.fill_val:
             return full_estimation
 
-        # Raise error if fill-values should be handled in >=3 dims
-        if self.grid_dim >= 3:
-            raise ValueError(
-                "Fill-value handling only supported in up to two grid dimensions."
-            )
-
         # Early exit if a nearest neighbor estimation would be overwritten
+        # Complex setup is needed to catch settings where the user mixes approaches and
+        # e.g. uses nearest neighbors for extrapolation and an actual interpolation otherwise
         if self.grid_dim == 1:
             if (
                 (target_point < self.grid_points.min())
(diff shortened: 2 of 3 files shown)


docs/changes/282.feature.rst
```diff
@@ -0,0 +1 @@
+Add simple combinatoric fill-value handling for RAD_MAX estimation.
```
maxnoe, line 1: "simple"? ....

pyirf/interpolation/component_estimators.py
```diff
@@ -540,8 +563,101 @@ def __call__(self, target_point):
...
+                (target_point < self.grid_points.min())
+                or (target_point > self.grid_points.max())
+            ) and issubclass(self.extrapolator.__class__, BaseNearestNeighborSearcher):
+                return full_estimation
+            elif issubclass(self.interpolator.__class__, BaseNearestNeighborSearcher):
+                return full_estimation
+        elif self.grid_dim == 2:
+            target_simplex = self.triangulation.find_simplex(target_point)
+
+            if (target_simplex == -1) and issubclass(
```
maxnoe, line 593: You are checking for `BaseNearestNeighborSearcher` here twice, only the check on `target_simplex` is added, but then you do the same in both cases.

So just checking for `issubclass(BaseNearestNeighborSearcher` should be enough?

pyirf/interpolation/component_estimators.py
```diff
@@ -540,8 +563,101 @@ def __call__(self, target_point):
             effective areas. For RAD_MAX_2D of shape (n_energy_bins, n_fov_offset_bins)
 
         """
+        # First, construct estimation without handling fill-values
+        full_estimation = super().__call__(target_point)
+        # Safeguard against extreme extrapolation cases
+        full_estimation[full_estimation < 0] = 0
```
maxnoe, line 569: `np.clip(full_estimation, 0, None, out=full_estimation)`

pyirf/interpolation/component_estimators.py
```diff
@@ -540,8 +563,101 @@ def __call__(self, target_point):
...
+
+        # Early exit if fill_values should not be handled
+        if not self.fill_val:
+            return full_estimation
+
+        # Raise error if fill-values should be handled in >=3 dims
+        if self.grid_dim >= 3:
```
maxnoe, line 576: Shouldn't this error be already raised in `__init__`, not just at `__call__`?

pyirf/interpolation/component_estimators.py
```diff
@@ -540,8 +563,101 @@ def __call__(self, target_point):
...
+            raise ValueError(
+                "Fill-value handling only supported in up to two grid dimensions."
+            )
+
+        # Early exit if a nearest neighbor estimation would be overwritten
```
maxnoe, line 581: Why can't we just check for nearest neighbor? Why this complex setup?

RuneDominik: That setup is to catch cases where the user decided to mix approaches and use e.g. NearestNeighbor for interpolation but an actual extrapolator for extrapolation or vice-versa. I'll extend the comment in code to make this more clear for the future.

pyirf/interpolation/component_estimators.py
RuneDominik: `target_simplex == -1` means `target_point` outside grid, thus this checks for cases, where extrapolation will actually be applied and only in these cases the `issubclass` check on `self.extrapolator.__class__` yields the needed information. In the second case, thus `target_simplex != -1`, this needs to check `self.interpolator.__class__`.

Basically, this accounts for the possibility that the user mixes and uses NearestNeighbor for extrapolation and actual interpolation otherwise (and vice-versa) at the same time.

maxnoe (approved):

PR #287, opened 2024-05-28
Update .mailmap

HealthyPear:
Not sure if GitHub respects the .mailmap file, but I lost access to the old email, and I do not seem to appear in the list of contributors, so I prefer to point to my "institute-agnostic" email...


maxnoe:
Github doesn't respect mailmap I think for the contributors graph.

But you should be able to have the email as "secondary email" in your account

maxnoe (approved):

PR #290, opened 2024-09-30
Multiple quantiles for angular_resolution

LukasBeiske:
This makes it possible to pass multiple values to the `quantile` argument of `pyirf.benchmarks.angular_resolution`. 
If done so, all the quantiles will be calculated and if only one quantile is passed (default), the output is the same as before.

### DIFF
--- a/docs/changes/290.feature.rst
+++ b/docs/changes/290.feature.rst
@@ -1,0 +1,2 @@
+Make it possible to pass multiple quantiles to ``pyirf.benchmarks.angular_resolution``, calculating all of them.
+If only one quantile is passed, the output is the same as before.
--- a/docs/changes/290.feature.rst
+++ b/docs/changes/290.feature.rst
@@ -1,1 +1,0 @@
-Make it possible to pass multiple quantiles to ``pyirf.benchmarks.angular_resolution``, calculating all of them.
--- a/docs/changes/290.feature.rst
+++ b/docs/changes/290.feature.rst
@@ -1,2 +1,1 @@
 Make it possible to pass multiple quantiles to ``pyirf.benchmarks.angular_resolution``, calculating all of them.
-If only one quantile is passed, the output is the same as before.
--- a/pyirf/benchmarks/angular_resolution.py
+++ b/pyirf/benchmarks/angular_resolution.py
@@ -10,15 +10,18 @@ ONE_SIGMA_QUANTILE = norm.cdf(1) - norm.cdf(-1)
 
 
 def angular_resolution(
-    events, energy_bins,
+    events,
+    energy_bins,
     energy_type="true",
-    quantile=ONE_SIGMA_QUANTILE,
+    quantile=[ONE_SIGMA_QUANTILE],
 ):
     """
     Calculate the angular resolution.
 
     This implementation corresponds to the 68% containment of the angular
     distance distribution.
+
+    Passing a list of quantiles results in all the quantiles being calculated.
 
     Parameters
     ----------
@@ -29,8 +32,8 @@ def angular_resolution(
     energy_type: str
         Either "true" or "reco" energy.
         Default is "true".
-    quantile : float
-        Which quantile to use for the angular resolution,
+    quantile : list(float)
+        Which quantile(s) to use for the angular resolution,
         by default, the containment of the 1-sigma region
         of the normal distribution (~68%) is used.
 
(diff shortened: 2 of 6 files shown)


pyirf/benchmarks/angular_resolution.py
```diff
@@ -10,16 +10,19 @@
 
 
 def angular_resolution(
-    events, energy_bins,
+    events,
+    energy_bins,
     energy_type="true",
-    quantile=ONE_SIGMA_QUANTILE,
+    quantile=[ONE_SIGMA_QUANTILE],
```
maxnoe, line 16: Please don't use mutable default arguments

pyirf/benchmarks/angular_resolution.py
```diff
@@ -52,8 +55,17 @@ def angular_resolution(
     result[f"{energy_key}_center"] = 0.5 * (energy_bins[:-1] + energy_bins[1:])
     result["n_events"] = 0
 
-    key = "angular_resolution"
-    result[key] = u.Quantity(np.nan, table["theta"].unit)
+    # keep backwards compatibility
+    if not isinstance(quantile, list) and not isinstance(quantile, np.ndarray):
```
maxnoe, line 59: `isinstance` supports tuples for testing multiple classes:
```
isinstance(quantile, (list, ndarray))
```
but better would probably just to test for `Sequence`

pyirf/benchmarks/angular_resolution.py
```diff
@@ -52,8 +56,16 @@ def angular_resolution(
     result[f"{energy_key}_center"] = 0.5 * (energy_bins[:-1] + energy_bins[1:])
     result["n_events"] = 0
 
-    key = "angular_resolution"
-    result[key] = u.Quantity(np.nan, table["theta"].unit)
+    if not isinstance(quantile, Sequence):
+        quantile = [quantile]
+
+    if len(quantile) < 2:
+        keys = ["angular_resolution"]
```
kosack, line 63: I think it's somewhat confusing to change the naming of the output key based on the number of quantiles. Just having a single valuye doesn't mean that that value is what should be defined as "angular resolution" (which is often defined as either 68% containment or 90% containment in different cases).  Wouldn't it be better to just always call them _containment_quantile_XX_ in all cases?  or if you really want to be specific, something like `angular_resolution_{value*100:.0f}pct_containment`.  That way there is no surprise to the user that the column names change.  

I guess the issue is that somewhere else, the code looks for a magic column named "angular_resolution", but then it's also better to make that column name configurable, and have it default to the CTAO standard using the more explicit column naming.

LukasBeiske: Good point, being consistent with the column naming will avoid unnecessary confusion, I'm not sure anymore why I didn't do it like this initially...
Within pyirf only one of the tutorials accesses the column name "angular_resolution", so I'll just change it there accordingly.

pyirf/benchmarks/angular_resolution.py
```diff
@@ -52,8 +56,13 @@ def angular_resolution(
     result[f"{energy_key}_center"] = 0.5 * (energy_bins[:-1] + energy_bins[1:])
     result["n_events"] = 0
 
-    key = "angular_resolution"
-    result[key] = u.Quantity(np.nan, table["theta"].unit)
+    if not isinstance(quantile, Sequence):
+        quantile = [quantile]
+
+    keys = [f"containment_quantile_{value * 100:.0f}" for value in quantile]
```
maxnoe, line 62: I'd prefer `angular_resolution_{}`

pyirf/benchmarks/angular_resolution.py
```diff
@@ -63,6 +72,8 @@ def angular_resolution(
     # use groupby operations to calculate the percentile in each bin
     by_bin = table[valid].group_by(bin_index[valid])
     for bin_idx, group in zip(by_bin.groups.keys, by_bin.groups):
-        result[key][bin_idx] = np.nanquantile(group["theta"], quantile)
-        result["n_events"][bin_idx] = len(group)
+        for key, value in zip(keys, quantile):
+            result[key][bin_idx] = np.nanquantile(group["theta"], value)
```
maxnoe, line 76: `np.nanquantile` can compute for multiple quantiles at the same time. Since sorting the input values is the most expensive operation, this is much faster than doing it one-by-one.

This was the main reason to add support for multiple quantiles to the function itself instead of calling it multiple times.

maxnoe (changes requested):

docs/changes/290.feature.rst
```diff
@@ -0,0 +1 @@
+Make it possible to pass multiple quantiles to ``pyirf.benchmarks.angular_resolution``, calculating all of them.
```
maxnoe, line 1: This is an API change now, as the column names of the table have changed

maxnoe (approved):

PR #295, opened 2025-01-20
Allow LogParabola to be defined using natural log

HealthyPear:
In some experiments or software the definition of a LogParabola spectral model might brequire to use the natural logarithm instead of the one in base 10.

This PR allows to instantiate such a model with the addition of a new class attribute analog to the one used by [gammapy.modeling.models.LogParabolaSpectralModel](https://docs.gammapy.org/1.3/api/gammapy.modeling.models.LogParabolaSpectralModel.html)


pyirf/spectral.py
```diff
@@ -195,25 +195,32 @@ class LogParabola:
         :math:`\beta`
     e_ref: astropy.units.Quantity[energy]
         :math:`E_\text{ref}`
+    from_log10: bool
+        If True, compute the energy ration in the exponent factor
+        using base 10, else use natural logarithm.
```
morcuended, line 200: Maybe here say that default it base 10 log

maxnoe: I don't particularly like the keyword name here, why not call this "log_base" or just pass the log function?

E.g. `log=np.log10` vs. `log=np.log`.

maxnoe: Ah, I think you took the name from https://docs.gammapy.org/0.19/api/gammapy.modeling.models.LogParabolaSpectralModel.html#gammapy.modeling.models.LogParabolaSpectralModel.from_log10


But this is not a keyword, it's an alternative constructor, that's why it starts with `from_`.

HealthyPear: yes, the idea was to keep a nomenclature analog to `gammapy` which is what people tends to use more lately

maxnoe: But it doesn't make sense to use the name of a class method for the name of a keyword

maxnoe: You could implement the same (or rather the inverse) logic as gammapy, e.g. offer a `from_ln` method that converts the parameters to base 10. 

Or you change the name to `log`, maybe with an enum like this:

```
class LogType(Enum):
    LOG10 = auto()
    LN = auto()

class LogParabola:
    def __init__(self, ..., log=LogParabolaBase.LOG10)

```

maxnoe (changes requested):
See Inline comment

PR #300, opened 2025-04-22
Add asymmetric energy dispersion/migration support

luca-dib:
This adds functions for calculating the asymmetric energy dispersion and migration matricies to `pyirf.irf.energy_dispersion` and creating FITS table HDUs for the asymmetric energy dispersion to `pyirf.io.gadf`.

The functions are separated into functions for a `theta` / `phi` and `lon` / `lat` FoV binning for the direct calculation of energy dispersion and migration and the HDUs:
- `energy_dispersion_asymmetric_polar`
- `energy_dispersion_asymmetric_lonlat`
- `energy_migration_matrix_asymmetric_polar`
- `energy_migration_matrix_asymmetric_lonlat`
- `create_energy_dispersion_asymmetric_polar_hdu`
- `create_energy_dispersion_asymmetric_lonlat_hdu`

Whereas the function for conversion from dispersion to migration `energy_dispersion_to_migration_asymmetric` does not require this.


pyirf/io/gadf.py
```diff
@@ -202,6 +204,130 @@ def create_energy_dispersion_hdu(
...
+    true_energy_bins,
+    migration_bins,
+    fov_offset_bins,
+    fov_position_angle_bins,
+    point_like=True,
+    extname="EDISP",
+    **header_cards,
+):
+    """
+    Create a fits binary table HDU in GADF format for the asymmetric energy dispersion with polar FoV coordinates.
```
kosack, line 218: I have a small worry about the terminology: Since it's not obvious that "asymmetric" here means "non-radially-symmetric", i.e. with an azimuthal dependence, it would be good to explain that better.  Otherwise, i would think "asymmetric energy dispersion" is asymmetric in  true energy, which it is, but that is not what we are talking about here.    Maybe just use  similar terminology for describing spatial models as gammapy:  1D vs 2D, i.e.:
* 1D-polar: model is radially symmetric, i.e. no azimuthal dependance, _f(r)_
* 2D-polar: model depends on radius and azimuth, _f(r,phi)_
* 2D-cartesian: projected fov_lon and fov_lat dependance, _f(x,y)._

pyirf/io/gadf.py
```diff
@@ -202,6 +204,130 @@ def create_energy_dispersion_hdu(
...
+    Parameters
+    ----------
+    energy_dispersion: numpy.ndarray
+        Energy dispersion array, must have shape
+        (n_energy_bins, n_migra_bins, n_source_offset_bins)
+    true_energy_bins: astropy.units.Quantity[energy]
+        Bin edges in true energy
+    migration_bins: numpy.ndarray
+        Bin edges for the relative energy migration (``reco_energy / true_energy``)
+    fov_offset_bins: astropy.units.Quantity[angle]
```
kosack, line 293: Docstring doesn't match signature: cut and paste error?

kosack (changes requested):

pyirf/io/gadf.py
luca-dib: Thanks for your input!
Yes, I can see there certainly is some ambiguity in just calling it "asymmetric" and I agree that there might be a better way to express this.
Gammapy does not use the "nD" terminology to describe just the spatial axes though (at least for the IRFs), but includes the energy axis as well, e.g. 
- `EnergyDispersion2D` axes: `energy_true` and `offset`, `migra` (This implies to me the `migra` axis is ignored. Is this intentional or an inconsistency?)
- `PSF3D` axes: `energy_true`, `offset` and `rad`.

I definitely think we should be consistent in the naming convention and that there is probably value in also being consistent with gammapy in the future. Whatever way is decided on should be applied to the rest of the IRF package in my opinion, but the details warrant some discussion. For reference, in the non-radially symmetric effective area I also chose the "3D" naming convention, see #281.

luca-dib: > I think the "ND" should refer to the dimensions in the phase space in which an IRF component is stored, excluding intrinsic axes of the IRF component itself.

This also seems reasonable to me, however then the GADF PSF_TABLE breaks with this again, as you said of course.

I would propose that for this PR in accordance with the GADF I change the naming to "3D", however this leaves the PSF for which I would also soon want to create a PR. Is this worth an issue to discuss the naming etc.?

kosack: Yes, I agree with the dimensionality being the spatial+energy, so my comment above becomes just d+1. 

But you are right... for PSF you really need to describe both the internal (source coordinate) and external (field-of-view coordinate) dimensionality to be fully unambiguous.   Not sure the best way to name those... e.g. a model with a "2D" distribution of the PSF that changes with radius and azimuth in the FOV:   _PSF2D_3D_?, _2DPSF3D_?.  It gets ugly!  @bkhelifi let's try to get this right for VODF...

pyirf/io/gadf.py
```diff
@@ -202,6 +204,130 @@ def create_energy_dispersion_hdu(
...
+
+    # required header keywords
+    header = DEFAULT_HEADER.copy()
+    header["HDUCLAS1"] = "RESPONSE"
+    header["HDUCLAS2"] = "EDISP"
+    header["HDUCLAS3"] = "POINT-LIKE" if point_like else "FULL-ENCLOSURE"
+    header["HDUCLAS4"] = "EDISP_3D"
+    header["DATE"] = Time.now().utc.iso
+    idx = edisp.colnames.index("MATRIX") + 1
+    header[f"CREF{idx}"] = "(ENERG_LO:ENERG_HI,MIGRA_LO:MIGRA_HI,DETX_LO:DETX_HI,DETY_LO:DETY_HI)"
```
maxnoe, line 325: I'd put the migra axis last, see https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/178

luca-dib: I see the point, however in that case for consistency's sake the same should be applied to the 2D energy dispersion as well in my opinion. After checking this causes problems with the current tests for IO, specifically with the `test_cref` function. Will try to resolve these.

maxnoe: I agree in principle, but we are in a different situation with the 2d energy dispersion: we have it in GADF with the preferred axis order defined, so we should follow it. That this order is not the one we would prefer is an issue for GADF.

The 3D is new, so we should do what we think is most appropriate.

kosack (dismissed):

kosack:
Please rebase on main to re-trigger the CI

kosack (dismissed):

pyirf/io/gadf.py
```diff
@@ -202,6 +204,130 @@ def create_energy_dispersion_hdu(
...
+    edisp = QTable()
+    edisp["ENERG_LO"], edisp["ENERG_HI"] = binning.split_bin_lo_hi(true_energy_bins[np.newaxis, :].to(u.TeV))
+    edisp["DETX_LO"], edisp["DETX_HI"] = binning.split_bin_lo_hi(fov_longitude_bins[np.newaxis, :].to(u.deg))
+    edisp["DETY_LO"], edisp["DETY_HI"] = binning.split_bin_lo_hi(fov_latitude_bins[np.newaxis, :].to(u.deg))
+    edisp["MIGRA_LO"], edisp["MIGRA_HI"] = binning.split_bin_lo_hi(migration_bins[np.newaxis, :])
+    # transpose as FITS uses opposite dimension order
+    edisp["MATRIX"] = u.Quantity(energy_dispersion.T[np.newaxis, ...]).to(u.one)
+
+    # required header keywords
+    header = DEFAULT_HEADER.copy()
```
maxnoe: The default header adds the keywords saying that this is GADF version 0.2, however these new EDISP parametrizations are not defined in GADF 0.2 (nor 0.3), so we shouldn't claim GADF 0.2 here.

pyirf/irf/energy_dispersion.py
```diff
@@ -215,3 +472,86 @@ def energy_dispersion_to_migration(
         migration_matrix[idx, :, :] = y
 
     return migration_matrix
+
+
+def energy_dispersion_to_migration_3d(
```
maxnoe: Why do we need a separate 3d version of this? The operation only happens along the `migration` axis turning it into the reconstructed energy axis, so it should be the same for 2d and 3d?

luca-dib: I have to say when I first wrote the code I just did not think about this at all, but true, we really don't, it's just unnecessary duplication.

The only thing is that the axis handling is a tiny bit different due to the different ordering for the 3d case (as discussed [here](https://github.com/cta-observatory/pyirf/pull/300#discussion_r2121532414)). But I would say I'll just solve this by making a case distinction between 2d and 3d as in the `_normalize_hist` function.

pyirf/io/gadf.py
luca-dib: Good point. It also includes the pyirf version used to create the IRFs though, so I would leave this in. 

Should the GADF version anyways be updated to 0.3? From a quick glance at the changes between 0.2 and 0.3 I don't see a reason not to.

maxnoe: Yes, but that can be another PR independent of this one.

docs/changes/300.feature.rst
```diff
@@ -0,0 +1,11 @@
+Add functions to calculate asymmetric energy dispersion and energy migration matrices to ``pyirf.irf.energy_dispersion``, supporting both binning in ``true_source_fov_offset`` / ``true_source_fov_position_angle`` and ``true_source_fov_lon`` / ``true_source_fov_lat``.
+These functions are
+* ``energy_dispersion_asymmetric_polar``
```
maxnoe: I'd not call this `asymmetric`, maybe just `_3d_polar` or `_thetaphi`

docs/changes/300.feature.rst
```diff
@@ -0,0 +1,11 @@
...
+* ``energy_dispersion_asymmetric_lonlat``
+* ``energy_migration_matrix_asymmetric_polar``
+* ``energy_migration_matrix_asymmetric_lonlat``
+* ``energy_dispersion_to_migration_asymmetric``
```
maxnoe: This was removed I think

docs/changes/300.feature.rst
luca-dib: In that case I'd leave it as `_3d_polar` for now for consistency with the already implemented `effective_area_3d_` functions, and if it's desired/recommended to change the naming to be more explicit about the axes used in the future, put that in a separate PR for all implemented 3D IRF functions.

maxnoe (approved):

kosack (approved):

PR #316, opened 2026-07-09
Fix gammapy and astropy regions conflict

luca-dib:
Up to the latest gammapy release at the time of writing (v2.1) `gammapy.regions` imports the function `pixel_scale_angle_at_skycoord` from astropy's `regions` package, which was removed/renamed in regions v0.12, see [here](https://github.com/gammapy/gammapy/blob/v2.1/gammapy/utils/regions.py#L32) for the relevant gammapy code, [here](https://github.com/astropy/regions/pull/650/changes#diff-99103805d4632d7b7570731408cd3e572b723f361998388ba904bc96fc20aa96) for the change in regions. `pyirf` uses gammapy code that imports this in `pyirf.conftest`.

As this breaks the CI for completely unrelated PRs I suggest to constrain regions to <0.12 for now, until the issue is fixed in Gammapy, which should already be the next release.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,8 @@ dependencies = [
 
 [project.optional-dependencies]
 gammapy = [
-  "gammapy >=1.0,<2.1",
+  "gammapy >=1.0",
+  "regions <0.12"
 ]
 docs = [
     "sphinx",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
 
 [project.optional-dependencies]
 gammapy = [
-  "gammapy >=1.0",
+  "gammapy >=1.0,<2.1",
 ]
 docs = [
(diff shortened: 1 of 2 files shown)


luca-dib:
Tests are still failing due to issues described in #317.

kosack (changes requested):
Please add a comment mentioning why regions is fixed t <0.12, so we remember to unfix it later when the issue is solved in gammapy.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
 [project.optional-dependencies]
 gammapy = [
   "gammapy >=1.0",
+  "regions <0.12"
```
kosack:
```suggestion
  "regions <0.12"  # gammapy uses function removed in 0.12, remove when fixed upstream
```

kosack (approved):

PR #317, opened 2026-07-09
Fixed broken IRF interpolation due to removed numpy cross 2-component vector functionality

luca-dib:
Numpy removed the 2-component vector cross product functionality from `numpy.cross` after it has been deprecated for a while, which causes the `plumb_point_dist` function to fail in this case.
This will cause CI to fail in completely unrelated PRs.

This fixes the issue by using the fact that the cross product of 2 vectors `a` and `b` in 2D is simply the determinant of the 2x2 matrix composed of `a` and `b`.

Another option would have been appending 0 to both vectors before using `np.cross()`. The last component of the resulting vector is then equivalent to the 2D cross product. However after a quick check showed that this is ~3x slower than the determinant method, see attached screenshot.

Screenshot_20260709_115559

### DIFF
--- a/docs/changes/317.bugfix.rst
+++ b/docs/changes/317.bugfix.rst
@@ -1,0 +1,1 @@
+Changes `interpolation.utils.plumb_point_distance()` to fix `numpy.cross()` 2-component vector functionality removed in numpy>=2.5.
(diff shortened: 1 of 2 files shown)


luca-dib:
Tests still fail due to issues described in #316.

kosack (approved):

